PR #287, opened 2021-04-27
Feature/remove dummy perception publisher

yamacir-kit:
## Types of PR

- [ ] New Features
- [ ] Bugfix


openscenario/openscenario_interpreter/package.xml
```diff
@@ -9,7 +9,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>awapi_accessor</depend>
+  <depend>concealer</depend>
```
hakuturu583, line 12: I think this naming is good, but ambiguous.

hakuturu583 (approved):
I think this commit is good, but naming will be rethink later.

PR #318, opened 2021-05-26
Feature/integration with autoware.auto

danielm1405:
build with AUTOWARE_AUTO flag defined instead of AUTOWARE_ARCHITECTURE_PROPOSAL


danielm1405:
Step 1: builds and does not explode when ran with AUTOWARE_AUTO flag instead of AUTOWARE_ARCHITECTURE_PROPOSAL

adamdbrw:
updated to rebase on current master

PR #328, opened 2021-06-04
Integration with AutowareAuto

adamdbrw:
## Types of PR

- [X] New Features

## Link to the issue

N/A

## Description

Work in Progress.

First iteration of changes (not in target design / architecture) to understand requirements and make a proof of concept for integration of scenario_simulator_v2 with AutowareAuto. 

## How to review this PR.

The PR is not for review yet (except for internal). 

Ifdef sections are meant to be replaced by a proper design once the changeset is operational and accepted.


test_runner/scenario_test_runner/test/scenario/autoware-simple.yaml
```diff
@@ -85,7 +85,7 @@ OpenSCENARIO:
                       dynamicsShape: step
                     SpeedActionTarget:
                       AbsoluteTargetSpeed:
-                        value: 2.778
+                        value: 0.0
```
adamdbrw, line 88: @pijaro Add comment here please

danielm1405: Done

yamacir-kit: The feature that gives Autoware a non-zero initial velocity is provided in case we want to test a scenario at cruising speed but cannot provide a map with straight lines that Autoware can accelerate sufficiently.
This is currently achieved in a very ad hoc way, but we are aware that it can adversely affect Autoware's planner.

If you have problems with the .Auto planner, you can temporarily turn this feature off.
In that case, instead of changing the scenario, fix it so that EgoEntity :: setTargetSpeed does nothing when it is .Auto.

danielm1405: Thank you for the explanation. We will change it as you suggested.

simulation/traffic_simulator/src/entity/ego_entity.cpp
```diff
@@ -333,7 +361,14 @@ void EgoEntity::setTargetSpeed(double value, bool)
 {
   Eigen::VectorXd v(5);
   {
+#ifdef AUTOWARE_ARCHITECTURE_PROPOSAL
     v << 0, 0, 0, value, 0;
+#endif
+#ifdef AUTOWARE_AUTO
+    // non-zero initial speed prevents behavioral planner from planning
+    // TODO (Robotec.ai) should we support non-zero initial speed?
```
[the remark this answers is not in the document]
adamdbrw: This would likely require some changes in Autoware Planner, not sure how deep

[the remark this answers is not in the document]
yamacir-kit: I agree with @adamdbrw -san.

adamdbrw:
For discussion: 

How to proceed after ifdef solution is deemed to work - with a design that would be easier to maintain

yamacir-kit:
@adamdbrw 

My thoughts on the discussion:

I am not thinking of an ArchitectureProposal-centric idea where we wait for Auto to have the same interface as ArchitectureProposal, but rather a gradual integration that respects both and proceeds step by step.

Previously we were not able to estimate the difference between the Auto and ArchitectureProposal simulation interfaces, but with this integration we know that it is only less than 1000 lines. I am happy to say that this is much smaller than expected.

I think the step that needs to be taken after this PR is merged is to change the process of switching between ArchitectureProposal and Auto to be done at runtime based on certain flags, instead of being done at compile time via ifdefs. Looking at the differences, this is quite possible.

I think it would be better to add the flag as a command line option in scenario_test_runner.py and propagate it to the inside of the simulator.
Of course, in the future, this option should be deprecated and scenario_test_runner.py should somehow determine whether it is ArchitectureProposal or Auto.

Next, it will be necessary to redesign the concealer by moving the VehicleModel etc. placed in EgoEntity to the concealer.
The redesigned concealer will choose to instantiate the `struct AutowareAuto` and `struct AutowareArchitectureProposal` derived from the common interface class `Autoware`.

Ideally, as Auto and ArchitectureProposal are integrated, the base class `Autoware` will become huge, and conversely, `struct AutowareAuto` and `struct AutowareArchitectureProposal` will become thin and finally disappear.

danielm1405:
@yamacir-kit 
We agree with the approach you proposed. We would be happy to help with the following steps.

danielm1405:
@yamacir-kit 
All checks passed except of **(lanelet2_extension_psim, galactic)** which seems unrelated to what this PR introduced (some connection error caused the failure).

external/concealer/include/concealer/autoware_def.hpp
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CONCEALER__AUTOWARE_DEF_HPP_
+#define CONCEALER__AUTOWARE_DEF_HPP_
+
+#define AUTOWARE_AUTO
```
yamacir-kit, line 18: NOTE: This code does not support both Auto and ArchitectureProposal at the same time.

Changing the default Autoware version to Auto will affect the existing Tier IV CI / CD pipeline (using scenario_simulator_v2 for testing).
I want ArchitectureProposal to be the default until dynamic Autoware architecture selection is supported.

external/concealer/src/execute.cpp
```diff
@@ -45,4 +51,27 @@ int execute(const std::vector<std::string> & f_xs)
 
   return ::execvp(argv[0], argv.data());
 }
+
+void sudokill(pid_t process_id)
+{
+  char * process_str;
```
yamacir-kit, line 57: Rather than a raw pointer, I think that is cool better to use the unique_ptr.

external/concealer/src/autoware.cpp
```diff
@@ -30,66 +33,17 @@ namespace concealer
...
-  AUTOWARE_INFO_STREAM("Shutting down Autoware: (3/3) Waiting for Autoware to be exited.");
-  {
-    int status = 0;
-
-    if (waitpid(process_id, &status, 0) < 0) {
-      AUTOWARE_SYSTEM_ERROR("waitpid");
-      std::exit(EXIT_FAILURE);
-    }
+    sudokill(process_id);
   }
```
yamacir-kit, line 47: The ugly code before the change played an important role.

First, if you simply kill the process that manages Autoware (`python3 ros2 launch ...`), the nodes launched by `python3 ros2 launch ...` will become zombies.

When ros2 launch receives a SIGINT, it sends a SIGINT to all the nodes it manages, and if it does not finish within 5 seconds, it transitions to a SIGKILL send.

However, if any node segfaults on termination, the ros2 launch process will never terminate because ros2 launch cannot determine if the node exited after SIGKILL.

(Sadly, SEGV at the end of Autoware happens quite often)

With the above in mind, you need to take the following steps to ensure that Autoware is shut down.

1. `python3 ros2 launch ...` Send SIGINT to the process
2. Wait for `ros2 launch` to send SIGKILL to the child process (= wait 5 seconds) and then send SIGKILL to the` ros2 launch` process itself.
3. Confirm the end of the `ros2 launch` process

Did these steps cause problems for Autoware.Auto?

yamacir-kit (changes requested):

yamacir-kit:
I have confirmed that there is no effect on the operation of ArchitectureProposal.
I'm going to check the operation with .Auto.

yamacir-kit:
I have checked that scenario_simulator_v2 and Autoware.Auto work fine in ADE.

external/concealer/src/autoware.cpp
WJaworskiRobotec: Actually the code that was in the destructor of Autoware before was not working properly for Autoware.Auto - it was only killing launch process, without proper shutdown of child processes (planning, perception etc. nodes). Thanks to replacement to sudokill method it now works properly - kills launch process, as well as all child processes. We can still refactor it in similar way that it was before if you prefer - basically sudokill method is replacement of sending SIGINT to ros2 launch.

external/concealer/src/execute.cpp
danielm1405: I refactored `sudokill` and it is no longer an issue.

external/concealer/src/execute.cpp
```diff
@@ -45,4 +51,27 @@ int execute(const std::vector<std::string> & f_xs)
...
+  pid_t pid;
+
+  if (asprintf(&process_str, "%d", process_id) < 0)
+    std::cout << std::system_error(errno, std::system_category()).what() << std::endl;
+
+  pid = fork();
+
+  switch (pid) {
+    case -1:
+      std::cout << std::system_error(errno, std::system_category()).what() << std::endl;
```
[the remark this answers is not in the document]
danielm1405: It has been fixed, thank you.

external/concealer/src/autoware.cpp
WJaworskiRobotec: It's been refactor to have all the logic you implemented before, but to work with Autoware.Auto as well.

danielm1405:
@yamacir-kit, we addressed all the comments. Is there anything else you would like us to improve in this PR?

external/concealer/src/autoware.cpp
yamacir-kit: I tried to run the refactored code with ArchitectureProposal, but sudokill still did not work properly with ArchitectureProposal.

I feel that we are facing a very complicated problem. However, since we have found code that works for both ArchitectureProposal and Auto, we think it is not worth the time to analyze this problem.

Therefore, we will postpone solving this problem.
In other words, make the destructor branch with an ifdef, and modify ArchitectureProposal to use the original code and Auto to use sudokill to do the shutdown.

WJaworskiRobotec: Do I understand correctly that previous simplified version with sudokill worked both for AutowareAuto and ArchitectureProposal, and new refactored version supports only AutowareAuto ?

hakuturu583:
Great contribution!
This is a very big and important feature, so please add it to the release note.
https://github.com/tier4/scenario_simulator_v2/blob/master/docs/ReleaseNotes.md

external/concealer/src/autoware.cpp
danielm1405: @yamacir-kit I addressed your comment. I branched killing Autoware for AAP (original code) and AA (`sudokill`) with `ifdef`s.

yamacir-kit: @WJaworskiRobotec

> Do I understand correctly that previous simplified version with sudokill worked both for AutowareAuto and ArchitectureProposal, and new refactored version supports only AutowareAuto ?

No, this is actually the case.

- Original code: Works correctly with AAP, but does not work properly with AA
- sudokill (first version): Doesn't work properly with AAP, but works correctly with AA
- sudokill (refactored version): Doesn't work properly with AAP, but works correctly with AA

yamacir-kit:
**I've checked that the code works as expected in both AAP and AA!**

Finally, list the features you have added in the `Difference between the latest release and master` section of the` docs / ReleaseNotes.md`.

I will merge as soon as I confirm the completion of the addition.

danielm1405:
@yamacir-kit
I updated `docs/ReleaseNotes.md`

yamacir-kit (approved):

yamacir-kit:
LGTM

PR #462, opened 2021-08-23
initialize current_time_ with negative value 

danielm1405:
Initialize `current_time_` with 0 so `setTargetSpeed` and `setTargetVelocity` are deterministic. 

Currently, half of the times call of the `setTargetSpeed` or `setTargetVelocity` results in error:
```
You cannot set entity status to the ego vehicle name:ego after starting scenario.
```
as half of the times `getCurrentTime()` returns positive value from randomly initialized double.

## Types of PR

- [ ] New Features
- [X] Bugfix


hakuturu583 (approved):
Thank you for fixing problem!

hakuturu583:
I will add your contribution to the release note.
https://github.com/tier4/scenario_simulator_v2/pull/463

PR #480, opened 2021-08-30
Remove unreachable return in API::spawn

prybicki:
## Types of PR

- [ ] New Features
- [x] Bugfix

### DIFF
--- a/simulation/traffic_simulator/src/api/api.cpp
+++ b/simulation/traffic_simulator/src/api/api.cpp
@@ -130,7 +130,6 @@ bool API::spawn(
   simulation_interface::toProto(params, *req.mutable_parameters());
   spawn_misc_object_entity_client_.call(req, res);
   return res.result().success();
-  return true;
 }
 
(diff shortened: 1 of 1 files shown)


simulation/traffic_simulator/src/api/api.cpp
```diff
@@ -130,7 +130,6 @@ bool API::spawn(
   simulation_interface::toProto(params, *req.mutable_parameters());
   spawn_misc_object_entity_client_.call(req, res);
   return res.result().success();
-  return true;
```
hakuturu583, line 133: If the entity was failed to spawn in sensor simulator, it res.result().success() == false.
So, please remain this line.

```
return res.result().success();
```

hakuturu583 (changes requested):

simulation/traffic_simulator/src/api/api.cpp
hakuturu583: Oh! Sorry.
I missed.
Your code works right.

hakuturu583 (approved):
Thank you for finding my mistake.

PR #710, opened 2022-02-22
Ajd 331 optimization 2nd stage

danielm1405:
## Types of PR

- [ ] New Features
- [ ] Bugfix

## Description

This PR aims to limit the recalculation of the trajectories. Each action has access to the reference trajectory that is precalculated on `traffic_simulator` level and passed to the behavior tree nodes. To determine entity actions, distance from the trajectory to other objects (entities, stop lines, traffic lights, etc.) is calculated multiple times. Only a part of the trajectory is taken into the consideration during these calculations - this part will be called trimmed trajectory.

Previously, the trimmed trajectory was obtained in the following way:
* waypoints were calculated based on the reference trajectory and the horizon (2.)
* the trajectory was calculated as a spline using the waypoints as knots (3.)
Moreover, the trajectory was recalculated multiple times in `tick()` as distance calculation functions required waypoints as an argument. 

This PR changes this flow. Instead of recalculating the trajectory, the trajectory is trimmed to match the desired starting and ending points (using `getTrimmed` method). This is done only once in each `tick()` call.

optimization-2nd-stage drawio

To sum up the advantages:
* previously: a **costly** spline calculation was performed **multiple** times in each `tick()` call
* now: relatively **cheap** spline trimming is performed **once** in each `tick()` call


simulation/behavior_tree_plugin/src/vehicle/follow_lane_sequence/stop_at_traffic_light_action.cpp
```diff
@@ -102,13 +105,12 @@ BT::NodeStatus StopAtTrafficLightAction::tick()
   if (waypoints.waypoints.empty()) {
     return BT::NodeStatus::FAILURE;
   }
-  const auto spline = traffic_simulator::math::CatmullRomSpline(waypoints.waypoints);
   const auto distance_to_traffic_stop_line =
-    hdmap_utils->getDistanceToTrafficLightStopLine(route_lanelets, waypoints.waypoints);
+    hdmap_utils->getDistanceToTrafficLightStopLine(route_lanelets, *trajectory);
```
hakuturu583, line 109: Currently, *trajectory cannot be null pointer.
However, please do null check for future development.

simulation/behavior_tree_plugin/src/vehicle/follow_lane_sequence/stop_at_traffic_light_action.cpp
```diff
@@ -102,13 +105,12 @@ BT::NodeStatus StopAtTrafficLightAction::tick()
...
   if (!distance_to_traffic_stop_line) {
     return BT::NodeStatus::FAILURE;
   }
-  distance_to_stop_target_ = getDistanceToTrafficLightStopLine(route_lanelets, waypoints.waypoints);
+  distance_to_stop_target_ = getDistanceToTrafficLightStopLine(route_lanelets, *trajectory);
```
hakuturu583, line 113: Currently, *trajectory cannot be null pointer.
However, please do null check for future development.

simulation/traffic_simulator/include/traffic_simulator/math/hermite_curve.hpp
```diff
@@ -56,6 +56,7 @@ class HermiteCurve
   double getMaximum2DCurvature() const;
   double getLength(size_t num_points) const;
   double getLength() const { return length_; }
+  void setLength(double length) { length_ = length; }
```
hakuturu583, line 59: Length should be calculated in hermite curve class.
Why do you add setLength function?

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
```diff
@@ -228,9 +256,30 @@ CatmullRomSpline::CatmullRomSpline(const std::vector<geometry_msgs::msg::Point>
...
+{
+  curves_ = curves;
+
+  for (const auto & curve : curves_) {
+    length_list_.emplace_back(curve.getLength());
+    maximum_2d_curvatures_.emplace_back(curve.getMaximum2DCurvature());
+  }
+
+  length_list_.front() = staring_curve_length;
+  curves_.front().setLength(staring_curve_length);
```
hakuturu583, line 274: Why do you set length?

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
```diff
@@ -228,9 +256,30 @@ CatmullRomSpline::CatmullRomSpline(const std::vector<geometry_msgs::msg::Point>
...
+  length_list_.back() = ending_curve_length;
+  curves_.back().setLength(ending_curve_length);
```
hakuturu583, line 276: Why do you set length?

hakuturu583 (approved):
This approval was my mistake.
Sorry.
Please ignore this approval.

hakuturu583 (changes requested):
Also, please update Release Note

hakuturu583:
@danielm1405 It seems, this feature is huge change, so please write description more detail.
Please refer [this PR](https://github.com/tier4/scenario_simulator_v2/pull/729).

simulation/traffic_simulator/include/traffic_simulator/math/hermite_curve.hpp
danielm1405: This is needed for curve trimming. This PR aims to avoid spline recalculation because this operation is very costly. Instead, the `trajectory` is calculated as a section of `reference_trajectory`. To calculate the subspline (which is just a part of the spline) you need to determine which curves should be included in the subspline. The middle curves are taken as a whole, but the first one and the last one need to be trimmed (because the start point of the subspline can be in the middle of the curve and we need only a part of it to construct the subspline). The curves should not change their shapes, but only change their length. So `setLength` is used to shorten the curves without the need to recalculate them from scratch (because it is costly as well).

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
danielm1405: This is the trimming of the starting curve described [here](https://github.com/tier4/scenario_simulator_v2/pull/710#discussion_r842661253).

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
danielm1405: This is the trimming of the ending curve described [here](https://github.com/tier4/scenario_simulator_v2/pull/710#discussion_r842661253).

simulation/traffic_simulator/include/traffic_simulator/math/hermite_curve.hpp
hakuturu583: I see, but I think this function is dangerous because of this is a public function.
And also, this function does not changes starting point and ending point of hermite curve it breaks the hermite curve class.

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
hakuturu583: I see, but I think this function is dangerous because of this is a public function.
And also, this function does not changes starting point and ending point of hermite curve it breaks the hermite curve class.

simulation/traffic_simulator/src/math/catmull_rom_spline.cpp
hakuturu583: I see, but I think this function is dangerous because of this is a public function.
And also, this function does not changes starting point and ending point of hermite curve it breaks the hermite curve class.

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/math/hermite_curve.hpp
danielm1405: Yes, I agree this solution is not perfect. I will try to come up with a better idea. If you have any suggestions on how to implement hermite curve trimming I would really appreciate it.

hakuturu583: I think problem is setting length.
Can you use constructor of HermiteCurve class?

danielm1405: We discussed this problem internally. We agree that length setting is a problem. We came up with a solution: we will introduce the `CatmullRomSubspline` class. It stores a reference to the original `CatmullRomSpline`, `start_s` and `end_s`. It provides an interface for accessing the underlying spline as it was trimmed (as it would start in `start_s` and end in `end_s`). The underlying spline would not be changed in any way. Therefore, the dangerous functions `HermiteCurve::setLength` and `CatmullRomSpline::getSubspline` will no longer be needed and will be removed.

What do you think about this approach?

hakuturu583: Do you mean you trimmed spline in CatmullRomSubspline class not in HermiteCurve class?
And s value of the Hermite curve class is always from 0 to it's length?
If so, I think it is a good idea!

danielm1405: I believe it's even better - I do not trim the spline at all. I use the `CatmullRomSubspline` class to access the spline as it was trimmed, but the underlying spline is not modified at all.

hakuturu583:
@danielm1405 Please tell me why you need CatmullRomInterface class?

simulation/traffic_simulator/include/traffic_simulator/math/catmull_rom_interface.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <geometry_msgs/msg/point.hpp>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace traffic_simulator
+{
+namespace math
+{
+class CatmullRomInterface
```
hakuturu583, line 29: I think it's naming is too ambiguous.
I think it is a base class for catmull rom spline curve.
How about rename to CatmullRomSplineBase or CatmullRomSplineInterface class.

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/math/catmull_rom_interface.hpp
danielm1405: Ok, I renamed `CatmullRomInterface` to `CatmullRomSplineInterface`

hakuturu583 (approved):

PR #713, opened 2022-02-25
Fix/interpreter/traffic signal state action

yamacir-kit:
## Types of PR

- [ ] New Features
- [x] Bugfix

## Description

- Fix `TrafficSignalStateAction::start` to not to throw error if given state is valid
- Update syntax `TrafficSignalState` to reset both of color and arrow if `none` is specified


hakuturu583 (approved):
LGTM

PR #727, opened 2022-03-11
Feature/interpreter/reader

yamacir-kit:
## Types of PR

- [x] New Features
- [ ] Bugfix

### DIFF
--- a/.github/workflows/BuildIsolate.yaml
+++ b/.github/workflows/BuildIsolate.yaml
@@ -25,8 +25,9 @@ jobs:
             kashiwanoha_map,
             lanelet2_extension_psim,
             lanelet2_matching,
-            openscenario_interpreter_msgs,
             openscenario_interpreter,
+            openscenario_interpreter_example,
+            openscenario_interpreter_msgs,
             openscenario_msgs,
             openscenario_utility,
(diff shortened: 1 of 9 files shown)


hakuturu583 (changes requested):
Code looks good to me.
Please add openscenario_interpreter_example package to the [depends of scenario_simulator_v2 package](https://github.com/tier4/scenario_simulator_v2/blob/master/scenario_simulator_v2/package.xml). 

And also, please add openscenario_interpreter_example package
to the [build_isolate action](https://github.com/tier4/scenario_simulator_v2/blob/48e42d8b2cb55eb7023318c703809a4fa3f5c928/.github/workflows/BuildIsolate.yaml#L20-L43).

yamacir-kit:
scenario_simulator_v2 does not depend on openscenario_interpreter_example.
If I add the dependency as you instructed, a circular reference will occur.

hakuturu583:
> scenario_simulator_v2 does not depend on openscenario_interpreter_example.
If I add the dependency as you instructed, a circular reference will occur.

I understand.
Thanks :+1:

hakuturu583 (approved):

PR #806, opened 2022-06-10
[Impl 2] Move non-instantaneous init actions into storyboard

HansRobo:
## Types of PR


HansRobo:
This is closed as this implementation was not adopted but #801 was

PR #848, opened 2022-07-31
Concealer service clients encapsulation; velocity limit setting fix

piotr-zyskowski-rai:
## Types of PR

- [ ] New Features
- [X] Bugfix

## Link to the issue
https://tier4.atlassian.net/browse/AJD-492

## Description
This PR fixes the issue, that the concealer failed to set the limit velocity. It was caused by the fact that the velocity adapter in autoware was not ready to accept the new limit velocity. This fix changes the behavior of autoware's concealer, so it would request the service several times before resigning.

Also, the changed code encapsulated this behavior into a `ServiceWithValidation` class. It is also applied to the `engage` service that also used the result validation behavior. The `cooperate commands` service that did not validate the service output and does not validate it in this code either but utilizes the `ServiceWithValidation` class.

## How to review this PR.
- Review the code
- Test whether scenarios in the task description pass


external/concealer/include/concealer/autoware_universe.hpp
```diff
@@ -40,6 +40,8 @@
 #include <tier4_rtc_msgs/msg/cooperate_status_array.hpp>
 #include <tier4_rtc_msgs/srv/cooperate_commands.hpp>
 
+#include "concealer/utility/service_with_validation.h"
```
yamacir-kit, line 43: If there is no reason why it must be double-quote, then use angled brackets to match other headers.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+// limitations under the License.
+
+#ifndef CONCEALER__SERVICE_WITH_VALIDATION_HPP_
+#define CONCEALER__SERVICE_WITH_VALIDATION_HPP_
+
+#include <chrono>
+#include <string>
+
+#include "concealer/autoware.hpp"
+#include "rclcpp/rclcpp.hpp"
```
yamacir-kit, line 22: If there is no reason why it must be double-quote, then use angled brackets to match other headers.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+
+namespace concealer
+{
+template <class T>
```
yamacir-kit, line 26: Use the `typename` keyword instead of the `class` keyword.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+class ServiceWithValidation
+{
+public:
+  ServiceWithValidation(std::string serviceName, Autoware & autoware)
```
yamacir-kit, line 30: The constructor should be explicit.
Even if it seems more convenient to be implicit, it is often detrimental in the long run.

yamacir-kit: Also, if there is no reason, do not copy the argument serviceName and use const reference to receive it.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  : serviceName(serviceName),
+    logger(autoware.get_logger()),
+    client(autoware.create_client<T>(serviceName, rmw_qos_profile_default)),
+    validation_rate(std::chrono::seconds(1))
+  {
+  }
+
+  void operator()(const typename T::Request::SharedPtr & request)
```
yamacir-kit, line 38: Use the trailing return type syntax for function return types.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    validateAvailability();
+    callWithTimeoutValidation(request);
+  }
+
+  void operator()(const typename T::Request::SharedPtr & request, size_t attempts_count)
```
yamacir-kit, line 44: Use the trailing return type syntax for function return types.
Also, use `std::size_t` for `size_t` if there is no particular reason.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    validateAvailability();
+    for (size_t attempt = 0; attempt < attempts_count; ++attempt, validation_rate.sleep()) {
+      std::optional<typename rclcpp::Client<T>::SharedFuture> future_opt =
+        callWithTimeoutValidation(request);
+      if (!future_opt) {
+        continue;
+      }
```
yamacir-kit, line 52: This early return is redundant.
Simply rewrite it to do the subsequent processing `if (future_opt)`.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    validateAvailability();
+    callWithTimeoutValidation(request);
+  }
+
+  void operator()(const typename T::Request::SharedPtr & request, size_t attempts_count)
+  {
+    validateAvailability();
+    for (size_t attempt = 0; attempt < attempts_count; ++attempt, validation_rate.sleep()) {
+      std::optional<typename rclcpp::Client<T>::SharedFuture> future_opt =
```
yamacir-kit, line 48: Do not use abbreviations such as `opt` unless your system imposes a limit such as a maximum length of characters.Also, the information given by a variable name of `future_opt` is not useful, since it is exactly the same information as the type.
A variable name should be a description of its use or role, not a type name, unless it is truly generic.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+        callWithTimeoutValidation(request);
+      if (!future_opt) {
+        continue;
+      }
+
+      auto result = future_opt->get();
```
yamacir-kit, line 54: Variables should be specified as const as possible.
It is also undesirable to copy data just to abbreviate `future_opt->get()` as `result`.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+          logger, serviceName << " service request has been accepted "
+                              << (result->status.message.empty()
+                                    ? "."
+                                    : " (" + result.get()->status.message + ")."));
+        break;
+      }
+      RCLCPP_ERROR_STREAM(
+        logger,
+        serviceName << " service request was accepted, but ResponseStatus is FAILURE "
+                    << (result->status.message.empty() ? "" : " (" + result->status.message + ")"));
```
yamacir-kit, line 66: Please specify the else clause.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    auto future = client->async_send_request(request);
+    if (future.wait_for(std::chrono::seconds(1)) != std::future_status::ready) {
+      RCLCPP_ERROR_STREAM(logger, serviceName << " service request has timed out.");
+      return std::nullopt;
+    }
+    return future;
+  }
+
+  std::string serviceName;
```
yamacir-kit, line 90: Use `lower_snake_case` for data members.
Also, this data member should be const.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+      }
+      RCLCPP_ERROR_STREAM(
+        logger,
+        serviceName << " service request was accepted, but ResponseStatus is FAILURE "
+                    << (result->status.message.empty() ? "" : " (" + result->status.message + ")"));
+    }
+  }
+
+private:
+  void validateAvailability()
```
yamacir-kit, line 71: Use the trailing return type syntax for function return types.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    while (!client->service_is_ready()) {
+      RCLCPP_INFO_STREAM(logger, serviceName << " service is not ready.");
+      validation_rate.sleep();
+    }
+  }
+
+  std::optional<typename rclcpp::Client<T>::SharedFuture> callWithTimeoutValidation(
+    const typename T::Request::SharedPtr & request)
```
yamacir-kit, line 80: Use the trailing return type syntax for function return types.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+  {
+    auto future = client->async_send_request(request);
+    if (future.wait_for(std::chrono::seconds(1)) != std::future_status::ready) {
```
yamacir-kit, line 83: Rewrite using the syntax "If statements with initializer", introduced in C++17.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+      RCLCPP_ERROR_STREAM(logger, serviceName << " service request has timed out.");
+      return std::nullopt;
+    }
+    return future;
```
yamacir-kit, line 87: Please specify the else clause.

external/concealer/src/autoware_universe.cpp
```diff
@@ -109,7 +109,7 @@ auto AutowareUniverse::engage() -> void
     waitForAutowareStateToBeDriving([this]() {
       auto request = std::make_shared<Engage::Request>();
       request->engage = true;
-      requestEngage(request);
+      requestEngage(request, 10);
```
yamacir-kit, line 112: Please add a comment as to why the number "10" is given.
In other words, please add the fact that "We attempt to resend the service up to 10 times, but this number of times was determined by heuristics, not for any technical reason.

external/concealer/src/autoware_universe.cpp
```diff
@@ -238,9 +238,11 @@ auto AutowareUniverse::sendSIGINT() -> void  //
 
 auto AutowareUniverse::setVelocityLimit(double velocity_limit) -> void
 {
-  auto request = std::make_shared<SetVelocityLimit::Request>();
-  request->velocity = velocity_limit;
-  requestSetVelocityLimit(request);
+  task_queue.delay([this, velocity_limit]() {
+    auto request = std::make_shared<SetVelocityLimit::Request>();
+    request->velocity = velocity_limit;
+    requestSetVelocityLimit(request, 10);
```
yamacir-kit, line 244: Please add a comment as to why the number "10" is given.
In other words, please add the fact that "We attempt to resend the service up to 10 times, but this number of times was determined by heuristics, not for any technical reason.

external/concealer/include/concealer/utility/service_with_validation.h
yamacir-kit: Also, change the variable `future_opt` so that it is not a local variable, but only scoped within an if statement (use the "if statements with initializer" syntax).

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,97 @@
...
+      }
+
+      auto result = future_opt->get();
+      if (result->status.code == tier4_external_api_msgs::msg::ResponseStatus::SUCCESS) {
+        RCLCPP_INFO_STREAM(
+          logger, serviceName << " service request has been accepted "
+                              << (result->status.message.empty()
+                                    ? "."
+                                    : " (" + result.get()->status.message + ")."));
+        break;
```
yamacir-kit, line 61: Rather than writing a process with an early return using break, the branch should be written carefully using the else clause.

external/concealer/include/concealer/utility/service_with_validation.h
yamacir-kit: (This is simply a matter of style preference, not a technicality)

yamacir-kit (changes requested):

yamacir-kit:
The processing and class design is perfect, but there are some things about the coding style that need to be fixed.

external/concealer/include/concealer/utility/service_with_validation.h
piotr-zyskowski-rai: I am sorry but I am not sure how to implement this. Could you elaborate on what would be the desirable code structure? :bow:

piotr-zyskowski-rai:
@yamacir-kit Thank you a lot! It was a very thoughtful review :bow: 

I applied almost all review comments. I only had [one additional question](https://github.com/tier4/scenario_simulator_v2/pull/848#discussion_r968647939)

external/concealer/include/concealer/utility/service_with_validation.h
yamacir-kit: I think loops with early return, continue, and break are prone to mistakes, so I think the pattern should be clearly shown by the exit condition of the for loop and the branch with the else clause.

For example, the break on line 61 in your code effectively leads to an exit from the function, but it's hard to read the intent at first glance.
Also, an error should be thrown if the loop completes attempts_count times and still does not succeed, but if you add that processing to the end of this function, break will also throw an error on success.

However, when I reviewed the code again, the part "the branch should be written carefully using the else clause" was incorrectly pointed out. I'm sorry.

After all, I think this function should be rewritten as:

```
auto operator()(const typename T::Request::SharedPtr & request, std::size_t attempts_count) -> void
{
  validateAvailability();

  for (auto attempt = attempts_count; 0 < attempt; --attempt)
  {
    if (auto future_opt = callWithTimeoutValidation(request);
        future_opt and
        future_opt->get()->status.code == tier4_external_api_msgs::msg::ResponseStatus::SUCCESS)
    {
      RCLCPP_INFO_STREAM(
        logger, serviceName << " service request has been accepted "
                            << (future_opt->get()->status.message.empty()
                                  ? "."
                                  : " (" + future_opt->get().get()->status.message + ")."));
      return;
    }
    else
    {
      RCLCPP_ERROR_STREAM(
        logger,
        serviceName << " service request was accepted, but ResponseStatus is FAILURE "
                    << (future_opt->get()->status.message.empty() ? "" : " (" + future_opt->get()->status.message + ")"));

      validation_rate.sleep();
    }
  }

  // THROW ERROR "requested the service attempts_count times, but was not successful."
}
```

yamacir-kit: After rewriting it, I felt that it is difficult to implement this function cleanly.
I don't think the code I showed above is pretty either.

piotr-zyskowski-rai: Thank you for the clarification. I have applied the return within the loop and error reporting, but I left the second if statement with the initializer of `service_call_status`, to avoid the repetition of `service_call_result->get()->status`.

Would the current structure be ok?

yamacir-kit: > Would the current structure be ok?
ok!

piotr-zyskowski-rai:
@yamacir-kit All conversation has been resolved. Are there some other changes I should introduce or could the PR be approved?

yamacir-kit:
> Are there some other changes I should introduce or could the PR be approved?

Approval.
However, to avoid accidental merging, I'll officially mark it as approve after confirming that there are no problems with the regression test by the Evaluator.

yamacir-kit:
The problem of Generate Documentation failing in GitHub Actions has already been resolved in master, so please merge origin/master into this branch.

piotr-zyskowski-rai:
Also, the CI documentation step seems to fail due to internal error I think
```
Error: Version 3.10.6 with arch x64 not found
```
I am not sure how to tackle this issue

yamacir-kit:
> I am not sure how to tackle this issue

The problem of Generate Documentation failing in GitHub Actions has already been resolved in master, so please merge origin/master into this branch.

bartek-kc:
Hello @yamacir-kit, I've tested the fix on behalf of @piotr-zyskowski-rai by running the test scenario in question >20 times locally and it passed during all the runs. I believe the PR is ready for re-review.

bartek-kc:
Hello @yamacir-kit, I've tested the fix on behalf of @piotr-zyskowski-rai by running the test scenario in question >20 times locally and it passed during all the runs. I believe the PR is ready for re-review.

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,99 @@
...
+      return future;
+    } else {
+      RCLCPP_ERROR_STREAM(logger, service_name << " service request has timed out.");
+      return std::nullopt;
+    }
+  }
+
+  const std::string service_name;
+  rclcpp::Logger logger;
+  typename rclcpp::Client<T>::SharedPtr client;
```
yamacir-kit, line 94: TODO @yamacir-kit 
- `typename rclcpp::Client<T>::SharedPtr` => `const typename rclcpp::Client<T>::SharedPtr`

external/concealer/include/concealer/utility/service_with_validation.h
```diff
@@ -0,0 +1,99 @@
...
+  rclcpp::WallRate validation_rate;
```
yamacir-kit, line 95: TODO @yamacir-kit 
- Provide a member function that modifies `validation_rate`.
- Make the setting value of `validation_rate` configurable by the constructor.
- Show the rationale for the initial value of `validation_rate`.

yamacir-kit (approved):

PR #884, opened 2022-09-21
Improve occupancy grid generation algorithm

shouth:
## Types of PR

- [ ] New Features
- [ ] Bugfix

## Description

This PR is to reduce occupancy grid map generation time. To achieve this goal, pixel traversal and grid filling is updated to use algorithms below.

- [A Fast Voxel Traversal Algorithm for Ray Tracing](https://www.researchgate.net/publication/2611491_A_Fast_Voxel_Traversal_Algorithm_for_Ray_Tracing)
- [Imos method](https://imoz.jp/algorithms/imos_method.html)

## How to review this PR.

This PR completely changes internal algorithm, so need to pay attention to corner cases.

## Others

I used only one test case to see how this PR works so this PR is not tested well.


shouth:
As far as I run this PR locally, occupancy grid generation takes 260**us** in a simple test case that contains only one NPC. More deitals are listed below.

- Resetting a builder ... avg 50**us**
- Adding a primitive ... avg 10**us**
- Building an occupancy grid ... avg 200**us**

Adding a primitve only takes 10**us**, so I think this PR is tolerant with adding many primitives compared to the old algorithm.

shouth:
I think algorithms in this PR are not best, but I do not come up with better solutions. Computer graphics algorithm such as shadow mapping may be applied, but I am not faimiliar with this field.

HansRobo:
To fix CI errors, it's nessesary to import changes from #885 after merging #885 to master!

shouth:
I changed some code about point manipulation, but this does not affect any behavior.

shouth:
Add some code to clip occupied area to fit into whole grid. I employed Boost.Geometry for convenience, but I can implement the algorithm by myself if we need to do so.

PR #902, opened 2022-10-26
Feature/simple noise simulator

kyoichi-sugahara:
## Types of PR

- [x] New Features
- [ ] Bugfix


kyoichi-sugahara:
@hakuturu583 @yamacir-kit 
I am in the process of implementing a noise generator, but I would appreciate it if you could confirm that I am going in the right direction.

PR #932, opened 2022-12-09
Feature/interpreter/alive monitoring

yamacir-kit:
## Types of PR

- [x] New Features
- [ ] Bugfix

## Description

Added new package `status_monitor`. The libraries contained in this package provide health checking functionality for any C++ program. The library is experimentally built into `openscenario_interpreter_node` and `simple_sensor_simulator_node`.

The health check is a function that allows the system running `scenario_simulator_v2` to monitor whether the components that make up `scenario_simulator_v2` are working properly. During its operation, `scenario_simulator_v2`, which supports the health check function, outputs information to the `/tmp/<component_name>_status.json` file every second to determine if the simulator is operating properly. the `good` element is false, the simulator is likely unresponsive.

Now, when Autoware Evaluator detects that the simulator supports the health check function, it continuously monitors the existence and content of `/tmp/<component_name>_status.json`, and one of the following conditions is met: force the simulator to shut down.

- The value of the entry `good` is `false`
- The value of the entry `unix_time` is much older than the current time (the status file has not been updated for more than the specified time)
- The status file does not exist

This feature is added to counter the problem of the simulator continuing to consume computing resources unnecessarily if it becomes unresponsive during simulation due to a bug such as an infinite loop.


HansRobo (approved):
Looks good to me with...
- [x] code reading
- [x] regression test
- [x] operation test with upper-level system ( by the author)

Special thanks for the detailed description!

PR #951, opened 2023-01-12
Fix/warnings

hakuturu583:
## Types of PR

- [ ] New Features
- [x] Bugfix

## Description

- suppress warnings from boost geometry
- suppress warnings from setuptools
- fix -Wrange-loop-construct warnings


.gitignore
```diff
@@ -12,6 +12,9 @@
 
 # external packages
 external/autoware_auto_msgs
+external/autoware_common/
+external/autoware_msgs/
+external/color_names/
 external/tier4_msgs
 external/zmqpp_vendor
```
HansRobo, line 19: I think this is simpler and less maintenance 
```suggestion
external/*
!external/lanelet2_matching
!external/concealer
```

HansRobo (changes requested):

.gitignore
hakuturu583: Fixed in https://github.com/tier4/scenario_simulator_v2/pull/951/commits/71e022ca98c3bea73375a8c5eea10d73db1a61cb

HansRobo: OK!

HansRobo (approved):
LGTM pending regression test

PR #971, opened 2023-02-24
Add `delay` implementation for `condition`

shouth:
## Types of PR

- [x] New Features
- [ ] Bugfix

## Description

Add delay implementation and condition edge implementation for condition.


openscenario/openscenario_interpreter/src/syntax/condition.cpp
```diff
@@ -43,10 +47,24 @@ Condition::Condition(const pugi::xml_node & node, Scope & scope)
...
+      return update_condition(std::function([](bool a, bool b) -> bool { return not a and b; }));
+
+    case ConditionEdge::risingOrFalling:
+      return update_condition(std::function([](bool a, bool b) -> bool { return a xor b; }));
+
+    case ConditionEdge::none:
+      return update_condition(std::function([](bool a) -> bool { return a; }));
+
+    case ConditionEdge::sticky:
+      return update_condition(std::function([this](bool a) -> bool { return current_value or a; }));
```
yamacir-kit, line 64: By using `!=` instead of `xor`, the explicit return type `bool` is unnecessary.

```suggestion
    case ConditionEdge::rising:
      return update_condition(std::function([](bool a, bool b) { return a and not b; }));

    case ConditionEdge::falling:
      return update_condition(std::function([](bool a, bool b) { return not a and b; }));

    case ConditionEdge::risingOrFalling:
      return update_condition(std::function([](bool a, bool b) { return a != b; }));

    case ConditionEdge::none:
      return update_condition(std::function([](bool a) { return a; }));

    case ConditionEdge::sticky:
      return update_condition(std::function([this](bool a) { return current_value or a; }));
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/condition.hpp
```diff
@@ -49,9 +57,36 @@ struct Condition : public ComplexType
...
+  std::list<History> histories;
+
+public:
   explicit Condition(const pugi::xml_node & node, Scope & scope);
 
   auto evaluate() -> Object;
+
+private:
+  template <class R, class... Args>
+  auto update_condition(std::function<R(Args...)> condition) -> Object
```
yamacir-kit, line 76: This function implicitly requires "a function whose argument types are any number of bool" (or "a function whose argument types are any number of types compatible with bool").
So it's nicer to express the type requirements as template parameter names.

```suggestion
  template <typename... Booleans>
  auto update_condition(std::function<bool(Booleans...)> condition)
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/condition.hpp
```diff
@@ -49,9 +57,36 @@ struct Condition : public ComplexType
...
+  {
+    histories.push_back({evaluateSimulationTime(), ComplexType::evaluate().as<Boolean>()});
+    if (auto iterator = std::find_if(
+          std::begin(histories), std::end(histories),
+          [this](const auto & entry) { return entry.time > histories.back().time - delay; });
+        std::ptrdiff_t(sizeof...(Args)) <= std::distance(std::begin(histories), iterator)) {
+      current_value = std::apply(condition, std::tuple{Args((--iterator)->result)...});
```
yamacir-kit, line 83: C-style casts should not be used.

```suggestion
        static_cast<std::ptrdiff_t>(sizeof...(Booleans)) <= std::distance(std::begin(histories), iterator)) {
      current_value = std::apply(condition, std::tuple(Booleans((--iterator)->result)...));
```

yamacir-kit (changes requested):

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/condition.hpp
shouth: Updated to use `static_cast` for `sizeof...(Args)`. But still need to use list-initialization `std::tuple{ ... }` because this line contains side-effect  ( `--iterator` ) and order of evaluation is significant. C++ computes expressions in order from left to right for list-initialization while order of evaluation for arguments in parenthesized initialization is *indeterminately sequenced* (arguments can be evaluated in order from left to right or from right to left).

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/condition.hpp
```diff
@@ -49,9 +57,37 @@ struct Condition : public ComplexType
...
+  template <typename... Booleans>
+  auto update_condition(std::function<bool(Booleans...)> condition) -> Object
+  {
+    histories.push_back({evaluateSimulationTime(), ComplexType::evaluate().as<Boolean>()});
+    if (auto iterator = std::find_if(
+          std::begin(histories), std::end(histories),
+          [this](const auto & entry) { return entry.time > histories.back().time - delay; });
+        static_cast<std::ptrdiff_t>(sizeof...(Booleans)) <=
+        std::distance(std::begin(histories), iterator)) {
+      current_value = std::apply(condition, std::tuple{Booleans((--iterator)->result)...});
```
shouth, line 84: This line depends on the order of evaluation of list initialization[^1] and cannot be replaced with direct initialization like `tuple(...)` whose order is *indeterminately sequenced*[^2]. This may be a confusing difference so it would be better to fold over the comma operator[^3].

[^1]: > 10 ) In [list-initialization](https://en.cppreference.com/w/cpp/language/list_initialization), every value computation and side effect of a given initializer clause is sequenced before every value computation and side effect associated with any initializer clause that follows it in the brace-enclosed comma-separated list of initializers.

[^2]: > 15 ) In a function call, value computations and side effects of the initialization of every parameter are indeterminately sequenced with respect to value computations and side effects of any other parameter. 
      > 17 ) Every expression in a comma-separated list of expressions in a parenthesized initializer is evaluated as if for a function call (indeterminately-sequenced)

[^3]: > 9 ) Every value computation and side effect of the first (left) argument of the built-in [comma operator](https://en.cppreference.com/w/cpp/language/operator_other#Built-in_comma_operator) , is sequenced before every value computation and side effect of the second (right) argument.
    
     https://en.cppreference.com/w/cpp/language/eval_order

```suggestion
      auto results = std::tuple<Booleans...>();
      std::apply([&](auto &... results) { ((results = (--iterator)->result), ...); }, results);
      current_value = std::apply(condition, results);
```

yamacir-kit: If the standard specifies the order of argument evaluation for list initialization, then there is nothing to worry about using list initialization.
Your suggestion of using the comma operator isn't bad, but it's still tricky either way.
That's why I prefer the original code you wrote.
In other words, the code below is OK.

```
      current_value = std::apply(condition, std::tuple{Booleans((--iterator)->result)...});
```

yamacir-kit (approved):

PR #996, opened 2023-04-27
Fix/get s value

hakuturu583:
## Types of PR

- [ ] New Features
- [x] Bugfix

## Description

- Change tolerance parameter of the polynomial solver (`std::numeric_limits<double>` => `1e-7`)
- Unify terms (root and solution) that refer to the same object used in variable names and comments
- Replaced root3 and _root3 functions that calculated cube roots with std::cbrt
- Add comments for the test cases and algorithms.
- Fix logic of the linear equation solver. For more detail, see also [this comment.](https://github.com/tier4/scenario_simulator_v2/pull/996#discussion_r1189335561)
- Corrected incorrectly used variable names (`root` => `discriminant`)
- Use push_back for std::vector<double> type in PolynomialSolver class and HermiteCurve class.

## How to review this PR.

Please check passing all CI test.


common/math/geometry/include/geometry/solver/polynomial_solver.hpp
```diff
@@ -84,6 +84,11 @@ class PolynomialSolver
  */
   double linearFunction(double a, double b, double t) const;
 
+  /**
+   * @brief Hard coded parameter, tolerance of calculation results of the PolynomialSolver
+   */
+  constexpr static double tolerance = 1e-7;
```
yamacir-kit, line 90: You should describe why this value was determined.
For example, "This value was determined by @hakuturu583 based on personal experience and not on any specific technical basis."

yamacir-kit, line 90: You should describe why this value was determined.
For example, "This value was determined by @hakuturu583 based on personal experience and not on any specific technical basis."
In particular, based on this change, you need to state why it is inappropriate in `std::numeric_limits<double>::epsilon`.

common/math/geometry/src/solver/polynomial_solver.cpp
```diff
@@ -105,15 +115,14 @@ std::vector<double> PolynomialSolver::solveCubicEquation(
...
-  const double eps = std::numeric_limits<double>::epsilon();
   double a2 = a * a;
   double q = (a2 - 3 * b) / 9;
   double r = (a * (2 * a2 - 9 * b) + 27 * c) / 54;
   // equation x^3 + q*x + r = 0
   double r2 = r * r;
   double q3 = q * q * q;
   double A, B;
-  if (r2 <= (q3 + eps)) {  //<<-- FIXED!
+  if (r2 <= (q3 + tolerance)) {  //<<-- FIXED!
```
yamacir-kit, line 125: What does the comment `<<-- FIXED!` mean? If you cannot explain a rational reason for this comment to exist, please delete it.

common/math/geometry/src/solver/polynomial_solver.cpp
```diff
@@ -38,9 +38,8 @@ double PolynomialSolver::quadraticFunction(double a, double b, double c, double
 std::vector<double> PolynomialSolver::solveLinearEquation(
   double a, double b, double min_value, double max_value) const
 {
-  constexpr double e = std::numeric_limits<double>::epsilon();
-  if (std::fabs(a) < e) {
-    if (std::fabs(b) < e) {
+  if (std::abs(a) < tolerance) {
```
yamacir-kit, line 41: This conditional expression is, after all, equivalent to
```
  if (std::abs(a) < tolerance) {
    return 0;
  }
```
Either you don't understand the code you're writing, or there's likely a bug lurking here.

common/math/geometry/src/solver/polynomial_solver.cpp
```diff
@@ -73,6 +76,10 @@ std::vector<double> PolynomialSolver::solveQuadraticEquation(
   for (const auto candidate : candidates) {
     if (min_value <= candidate && candidate <= max_value) {
       ret.emplace_back(candidate);
+    } else if (std::abs(candidate - max_value) < tolerance) {
+      ret.emplace_back(max_value);
+    } else if (std::abs(candidate - min_value) < tolerance) {
+      ret.emplace_back(min_value);
```
yamacir-kit, line 82: Why are you using emplace_back?
There is no reason to use emplace_back on a container whose element type is a scalar type.

yamacir-kit (changes requested):

common/math/geometry/src/solver/polynomial_solver.cpp
hakuturu583: Sorry, I deleted this comment because I can only assume that the comment was somehow left behind because its intent was unclear.

https://github.com/tier4/scenario_simulator_v2/pull/996/commits/6f96099d56e2a85b309793f48376872f14acb461

common/math/geometry/src/solver/polynomial_solver.cpp
hakuturu583: Change to push back in this commit.

https://github.com/tier4/scenario_simulator_v2/pull/996/commits/27921ff1a68898236e5a34e2f7a8a1b5b3cae242

common/math/geometry/src/solver/polynomial_solver.cpp
hakuturu583: I confirmed it was comes from We deleted comments that were unclear in intent, reviewed the process, and added comments to sections where the process was unclear.

common/math/geometry/src/solver/polynomial_solver.cpp
hakuturu583: I checked the formulation of the logic and I found that old version of the code returns the wrong value when the a and b value is small.
So, I changed the specification of the function.

hakuturu583: Before:
:o:  `ax + b = 0 (a=0, b≠0)` => return {} (it means, there is no solution.)
:x:  `ax + b = 0 (a=0, b=0)` and `(min_value <= 0 && 0 <= max_value) == false`=> return {} (it means, there is no solution. But actual, any rational number between min_value ~ max_value is a solution. )
:x: `ax + b = 0 (a=0, b=0)` and `(min_value <= 0 && 0 <= max_value) == true` => return {0} (it means, 0 is a solution.  But actual, any rational number between min_value ~ max_value is a solution. )

After:
:o: `ax + b = 0 (a=0, b≠0)` => return {} (it means, there is no solution.)
:o: `ax + b = 0 (a=0, b=0)` => throw SimulationError (Any rational number between min_value ~ max_value is a solution, so infinite numbers of solutions are exist.)

common/math/geometry/include/geometry/solver/polynomial_solver.hpp
hakuturu583: I described.

- Who decided this hard-coded parameter.
- Why the hard coded parameter should not std::numeric_limits<double>::epsilon.
- Why the hard coded parameter value is 1e-7.

On the comment.

common/math/geometry/src/solver/polynomial_solver.cpp
hakuturu583: use push_back for std::vector<double> type in PolynomialSolver class and HermiteCurve class.

common/math/geometry/src/solver/polynomial_solver.cpp
```diff
@@ -12,176 +12,212 @@
...
+                 });
+    }
+    /// @note No solutions are found.
+    return {};
+  };
+  return std::apply(solve_without_limit, tschirnhaus_transformation(a, b, c));
 }
 
-double PolynomialSolver::root3(double x) const
+auto PolynomialSolver::isEqual(double value0, double value1) const -> bool
```
yamacir-kit, line 218: The name `isEqual` is inappropriate, since this operation does not evaluate the two floating point numbers to be exactly equal. I suggest `isApproximatelyEqualTo` as a more appropriate name.

common/math/geometry/src/spline/hermite_curve.cpp
```diff
@@ -134,24 +135,42 @@ std::optional<double> HermiteCurve::getCollisionPointIn2D(
...
+       * @note Obtain a solution to the cubic equation ax^3 + bx^2 + cx + d = 0 that falls within the range [0, 1].
+       */
+      return solver_.solveCubicEquation(a, b, c, d, 0, 1);
+    }
+    /**
+    * @note PolynomialSolver::solveCubicEquation throws common::SimulationError when any x value can satisfy the equation, 
+    * so the beginning and end point of this curve can collide with the line segment.
+    * If search_backward = true, the line segment collisions at the end of the curve. So return 1.
+    * If search_backward = false, the line segment collisions at the start of the curve. So return 0.
+    */
```
yamacir-kit, line 151:
```suggestion
    /**
     * @note PolynomialSolver::solveCubicEquation throws common::SimulationError when any x value can satisfy the equation, 
     * so the beginning and end point of this curve can collide with the line segment.
     * If search_backward = true, the line segment collisions at the end of the curve. So return 1.
     * If search_backward = false, the line segment collisions at the start of the curve. So return 0.
     */
```

yamacir-kit (changes requested):

common/math/geometry/package.xml
```diff
@@ -16,6 +16,7 @@
   <depend>scenario_simulator_exception</depend>
   <depend>traffic_simulator_msgs</depend>
   <depend>tf2_geometry_msgs</depend>
+  <depend>boost</depend>
```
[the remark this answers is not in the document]
hakuturu583: Thanks, I sorted depend / test_depend tag in package.xml.

yamacir-kit (approved):

PR #1009, opened 2023-05-25
getStopLinesOnPath() changed from private to public

hrjp:
## Types of PR

- [ ] New Features
- [ ] Bugfix

## Description
Changed getStopLinesOnPath function from private to public for use in context_gamma_planner


hakuturu583:
If I check the source code, I get lanelet::ConstLineString3d as the return type, but what is this needed for?
If we only want to know the distance to the stop line for the center line only, wouldn't it be enough to know the lanelet_id of the stop line?

hakuturu583 (approved):
LGTM

Issue #1040, opened 2023-07-21
Brash-up communication protocol between sensor simulator and traffic simulator.


hakuturu583:
**Is your feature request related to a problem? Please describe.**
[Currently, sending simulation_api_schema::UpdateEntityStatusRequest for each entity.](https://github.com/tier4/scenario_simulator_v2/blob/794cd833078321c4edce1fc08d7b5ba0822dbae0/simulation/traffic_simulator/src/api/api.cpp#L293-L296)
It is not efficient way.

**Describe the solution you'd like**
We should use `repeated` field in protobuf schema and send as array.

PR #1058, opened 2023-07-31
ref(zeromq): restore repeated in UpdateEntityStatus

dmoszynski:
## Types of PR

- [ ] New Features
- [ ] Bugfix

## Link to the issue
#1040 
https://tier4.atlassian.net/browse/RJD-553
## Description

- I’ve analyzed the current implementation of `UpdateEntityStatusRequest`  usage and the previous one.

- Unfortunately, I haven't found a reason or advantage of removing repeated for the current implementation.

- The documentation of the assumptions doesn’t clearly indicate why such a decision was made. I didn't get any good reason from from the person responsible for the change.

-  It is true that for the current solution (no repeated) the update of each NPC (Non-Ego Entities) is done by a separate request - previously it was done on one.

- In order to restore `repeated`, I've developed a solution using 2 request calls, where: one request is for a status update for Ego and the other for all NPCs (Non-Ego Entities).
- In addition, I've optimized the `isEntityExists` method, as it had a big impact on call execution time.

## Others
I [ran evaluator](https://github.com/tier4/sim_evaluation_tools/issues/173) and got 2 negative results.
I've tested them and there is no regression, [here](https://tier4.atlassian.net/browse/RJD-553?focusedCommentId=132022) is the report confirming it.


hakuturu583:
Is this PR is related to #1040 ?

dmoszynski:
@hakuturu583 
That's right. After providing a green github CI I will run the evaluator - after I get positive results I will complete the PR description.

hakuturu583:
@dmoszynski This PR has a conflict, please resolve first.

dmoszynski:
@hakuturu583 
Indeed, the conflict must have arisen yesterday -  I did not see that. I've already solved it and tested a fresh build locally.

dmoszynski:
@hakuturu583 
Unfortunately, there is a problem with the master build after the last merge. After merging to this branch and resolving conflicts, I have the same issue: [master build actions](https://github.com/tier4/scenario_simulator_v2/actions/runs/5760667425/job/15617057170). I will wait for the problem to be solved.

simulation/traffic_simulator/src/api/api.cpp
```diff
@@ -243,22 +243,40 @@ bool API::updateTrafficLightsInSim()
...
-  if (res.result().success()) {
-    simulation_interface::toMsg(res.status().pose(), status_non_canonicalized.pose);
+  if (res.result().success() && req.status_size() == 1) {
+    assert(
+      res.status(0).name() == entity_name &&
+      "The entity name in response is different from the name in request!");
+    simulation_interface::toMsg(res.status(0).pose(), status_non_canonicalized.pose);
     simulation_interface::toMsg(
-      res.status().action_status(), status_non_canonicalized.action_status);
+      res.status(0).action_status(), status_non_canonicalized.action_status);
```
hakuturu583, line 279: Why do you hard-coded `0`?
Please remove hard-coded value if possible.
If you cannot remove this value, please leave doxygen comment and describe why the value should be `0`

simulation/traffic_simulator/src/api/api.cpp
```diff
@@ -243,22 +243,40 @@ bool API::updateTrafficLightsInSim()
...
+  simulation_interface::toProto(status_non_canonicalized, *req.add_status());
+
   simulation_api_schema::UpdateEntityStatusResponse res;
   zeromq_client_.call(req, res);
 
-  if (res.result().success()) {
-    simulation_interface::toMsg(res.status().pose(), status_non_canonicalized.pose);
+  if (res.result().success() && req.status_size() == 1) {
+    assert(
+      res.status(0).name() == entity_name &&
```
hakuturu583, line 275: Why do you hard-coded `0`?
Please remove hard-coded value if possible.
If you cannot remove this value, please leave doxygen comment and describe why the value should be `0`

simulation/traffic_simulator/src/api/api.cpp
```diff
@@ -243,22 +243,40 @@ bool API::updateTrafficLightsInSim()
...
-  simulation_interface::toProto(status_non_canonicalized, *req.mutable_status());
-  req.set_npc_logic_started(entity_manager_ptr_->isNpcLogicStarted());
+  simulation_interface::toProto(status_non_canonicalized, *req.add_status());
+
   simulation_api_schema::UpdateEntityStatusResponse res;
   zeromq_client_.call(req, res);
 
-  if (res.result().success()) {
-    simulation_interface::toMsg(res.status().pose(), status_non_canonicalized.pose);
+  if (res.result().success() && req.status_size() == 1) {
```
hakuturu583, line 273: Why do you hard-coded `1`?
Please remove hard-coded value if possible.
If you cannot remove this value, please leave doxygen comment and describe why the value should be `0`

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
```diff
@@ -171,6 +181,7 @@ void ScenarioSimulator::spawnVehicleEntity(
   }
   if (req.is_ego()) {
     ego_vehicles_.emplace_back(req.parameters());
+    entity_status_.insert({req.parameters().name(), simulation_api_schema::EntityStatus{}});
```
hakuturu583, line 184: Why do you initialize with initial value?
I think this value should be initialized from value `req`.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
```diff
@@ -184,6 +195,7 @@ void ScenarioSimulator::spawnVehicleEntity(
     ego_entity_simulation_->setInitialStatus(initial_status);
   } else {
     vehicles_.emplace_back(req.parameters());
+    entity_status_.insert({req.parameters().name(), simulation_api_schema::EntityStatus{}});
```
hakuturu583, line 198: Why do you initialize with initial value?
I think this value should be initialized from value `req`.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
```diff
@@ -195,6 +207,7 @@ void ScenarioSimulator::spawnPedestrianEntity(
   simulation_api_schema::SpawnPedestrianEntityResponse & res)
 {
   pedestrians_.emplace_back(req.parameters());
+  entity_status_.insert({req.parameters().name(), simulation_api_schema::EntityStatus{}});
```
hakuturu583, line 210: Why do you initialize with initial value?
I think this value should be initialized from value `req`.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
```diff
@@ -205,6 +218,7 @@ void ScenarioSimulator::spawnMiscObjectEntity(
   simulation_api_schema::SpawnMiscObjectEntityResponse & res)
 {
   misc_objects_.emplace_back(req.parameters());
+  entity_status_.insert({req.parameters().name(), simulation_api_schema::EntityStatus{}});
```
hakuturu583, line 221: Why do you initialize with initial value?
I think this value should be initialized from value `req`.

hakuturu583 (changes requested):

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: This method allows to call `UpdateEnityStatusRequest` for exactly one `Entity` (especially `Ego`) - for this reason I added a check if there is also only one status in the returned `Response` (check `req.status_size() == 1`) and perform operations only for this status (`res.status(0).name()`)

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: This method allows to call UpdateEnityStatusRequest for exactly one Entity (especially Ego) - for this reason I added a check if there is also only one status in the returned Response (check req.status_size() == 1) and perform operations only for this status (res.status(0).name())

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: This method allows to call UpdateEnityStatusRequest for exactly one Entity (especially Ego) - for this reason I added a check if there is also only one status in the returned Response (check req.status_size() == 1) and perform operations only for this status (res.status(0).name())

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: `entity_status_` is `std::map` for [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) with entity `name` as `key`. 
At the time of spawning, I add an entity to this map to only indicate that such an object exists, and leave its status empty because it will be updated. 
The values in this map are not used before update entities status. 
But of course I can develop the [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19)  initialization for each spawn operation.
([spawn requests](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L115-L121) don't have simple fields that can be rewritten to [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) , but I should be able to handle it)

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: `entity_status_` is `std::map` for [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) with entity `name` as `key`. 
At the time of spawning, I add an entity to this map to only indicate that such an object exists, and leave its status empty because it will be updated. 
The values in this map are not used before update entities status. 
But of course I can develop the [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19)  initialization for each spawn operation.
([spawn requests](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L115-L121) don't have simple fields that can be rewritten to [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) , but I should be able to handle it)

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: `entity_status_` is `std::map` for [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) with entity `name` as `key`. 
At the time of spawning, I add an entity to this map to only indicate that such an object exists, and leave its status empty because it will be updated. 
The values in this map are not used before update entities status. 
But of course I can develop the [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19)  initialization for each spawn operation.
([spawn requests](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L115-L121) don't have simple fields that can be rewritten to [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) , but I should be able to handle it)

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: `entity_status_` is `std::map` for [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) with entity `name` as `key`. 
At the time of spawning, I add an entity to this map to only indicate that such an object exists, and leave its status empty because it will be updated. 
The values in this map are not used before update entities status. 
But of course I can develop the [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19)  initialization for each spawn operation.
([spawn requests](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L115-L121) don't have simple fields that can be rewritten to [EntityStatus](https://github.com/tier4/scenario_simulator_v2/blob/11c3ea1998edeb80eae2eb43b9e4cccc7c3a9bb3/simulation/simulation_interface/proto/simulation_api_schema.proto#L12-L19) , but I should be able to handle it)

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: I improved the readability - added constexpr and provide a comment.

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: I improved the readability - added constexpr and provide a comment.

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: I improved the readability - added constexpr and provide a comment.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: I developed a method `insertEntitySpawned` Status that fills and adds a new entity_status_ during spawning.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: I developed a method `insertEntitySpawnedStatus` that fills and adds a new `entity_status_` during spawning.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: I developed a method `insertEntitySpawnedStatus` that fills and adds a new `entity_status_` during spawning.

simulation/simple_sensor_simulator/src/simple_sensor_simulator.cpp
dmoszynski: I developed a method `insertEntitySpawnedStatus` that fills and adds a new `entity_status_` during spawning.

simulation/traffic_simulator/src/api/api.cpp
```diff
@@ -232,19 +232,41 @@ bool API::updateTrafficLightsInSim()
...
-  simulation_interface::toProto(status_non_canonicalized, *req.mutable_status());
-  req.set_npc_logic_started(entity_manager_ptr_->isNpcLogicStarted());
-  if (auto res = zeromq_client_.call(req); res.result().success()) {
-    simulation_interface::toMsg(res.status().pose(), status_non_canonicalized.pose);
+  simulation_interface::toProto(status_non_canonicalized, *req.add_status());
+
+  simulation_api_schema::UpdateEntityStatusResponse res;
+  // The method updates a single entity status, so a single status is expected in response
+  if (auto res = zeromq_client_.call(req);
+      res.result().success() && req.status_size() == single_status_update) {
```
hakuturu583, line 262: When we use simple_sensor_simulator, `req.status_size()` should be 2, but the simulator such as AWSIM has a dynamics engine, so such simulators can change non-ego entity status.
So, we should check all updated entity status.

simulation/simulation_interface/proto/simulation_api_schema.proto
```diff
@@ -186,7 +186,7 @@ message UpdateEntityStatusRequest {
  **/
 message UpdateEntityStatusResponse {
   Result result = 1;                       // Result of [UpdateEntityStatusRequest](#UpdateEntityStatusRequest)
-  UpdatedEntityStatus status = 2;          // Updated entity status in sensor/dynamics simulator
+  repeated UpdatedEntityStatus status = 2;          // List of updated entity status in sensor/dynamics simulator
```
hakuturu583, line 189: Please format line.

hakuturu583 (changes requested):

simulation/traffic_simulator/src/api/api.cpp
dmoszynski: @hakuturu583 
Do you mean that even though there is only one status to update in the request, we can receive several statuses in the response (because they could change independently of the requests)?

hakuturu583: UpdateEntityStatusRequest send status of multiple entities.
I have determined that some simulator types may want to update EntityStatus during frame updates, independently of the traffic_simulator logic.
So, not only Ego entity but also non-Ego entity status can be updated by sensor simulator.

dmoszynski: @hakuturu583 
Thank you for your answer, please confirm which version best corresponds to the expected implementation:

#### Current
current drawio

#### First proposal
processed drawio

#### Second proposal
multiple drawio

### Received `UpdatedEntityStatus`
As I understand in both proposals, it is necessary to implement the processing of the received status (non-Ego entities) on the ss2 side.

Please let me know if the received status should be simply rewritten to `EntityManager`, or if a check should be implemented: if the received status is equal to the requested one.  Such a check seems to me to be essential.
I mean checking if the values in the received `Pose` are equal to the requested one - taking into account the `epsilon` value, e.g. `0.001` (does `ss2` have a parameter that can be used in such a check?).

Also, let me know what is the expected behavior when the expected values are not set - should an exception be thrown?

hakuturu583: Screenshot from 2023-08-25 11-14-10

I think this design is better because of overhead of the communication is minimum.

I think only updated entity status in `simple_sensor_simulator/AWSIM` should be packed into response.
If the requested status does not exists in response, (it can be checked by data field [name](https://github.com/tier4/scenario_simulator_v2/blob/408b00a4ffd3a8066cc777b72d40404f38fee141/simulation/simulation_interface/proto/traffic_simulator_msgs.proto#L75).) it means `No Operation` so do not update entity status in traffic_simulator.

hakuturu583: [Currently, non-ego entity does not it's status by returned response.](https://github.com/tier4/scenario_simulator_v2/blob/290ff1b9c2bab79b613ceb2aaa3acb0a1c7c55b6/simulation/traffic_simulator/src/api/api.cpp#L235-L246)
So, it should be also implemented.

hakuturu583: Currently, you implement update ego entity status before updating non-ego entity status.
I think we can do it at the same time.
Is there some problem about this plan?

dmoszynski:
@hakuturu583 

> Is there some problem about this plan?

This plan is like a very good approach, if we would like to process every `EntityStatus` received in `Response` (not just `Ego` as in the version from the current [main](https://github.com/tier4/scenario_simulator_v2/blob/408b00a4ffd3a8066cc777b72d40404f38fee141/simulation/traffic_simulator/src/api/api.cpp#L258-L268) branch).

Therefore,
as [requested](https://github.com/tier4/scenario_simulator_v2/pull/1058#discussion_r1305058876), I've implemented the processing of the `EntityStatus` received in the `Response` for each `Entity` - on the `traffic_simulator` side. I provided that any received `EntityStatus` is rewritten to `EntityManager` - I did not add the check I [mentioned](https://github.com/tier4/scenario_simulator_v2/pull/1058#discussion_r1301533894):
https://github.com/tier4/scenario_simulator_v2/blob/85601c0b387a5cf7503605e7cfb0d126389cca81/simulation/traffic_simulator/src/api/api.cpp#L246-L279

Due to the combination of updating all `EntityStatus` in one method, I also modified the `updateFrame` to the following form:
https://github.com/tier4/scenario_simulator_v2/blob/85601c0b387a5cf7503605e7cfb0d126389cca81/simulation/traffic_simulator/src/api/api.cpp#L281-L305

I have tested the changes locally on several scenarios, now I am waiting for the green github CI, after success I will launch the evaluator.

hakuturu583:
Thanks, I will waiting for end CI.

dmoszynski:
After merging from the master to this branch, the `FollowTrajectoryAction` test [(github CI)](https://github.com/tier4/scenario_simulator_v2/actions/runs/6026010079/job/16348002028#step:13:325) with the star path returns a failure.
It seems that as a result of the changes and merge, there is a problem where the `desired_acceleration` in `FollowTrajectoryAction` has a value of `-nan` - which throwing an exception.

While investigating the problem, I noticed that there is a difference in time (one step to be exact) during the execution of the next steps of the algorithm - between the master and the developed version.


The difference is due to the added call::
https://github.com/tier4/scenario_simulator_v2/blob/f3b0abc14875dc48fd7f12f9b1dbc64c44f6d703/simulation/traffic_simulator/src/api/api.cpp#L259

In which there is an override of the time existing in `EntityStatus` with the value of `EntityManager` internal `current_time_` (line 150).
https://github.com/tier4/scenario_simulator_v2/blob/bfd3344a828d24c4e22bfa7fab1b556a34bf91a5/simulation/traffic_simulator/src/entity/entity_manager.cpp#L143-L153

Unfortunately, after ending the update `EntityManager` internal `current_time` remains outdated - `Entities` statuses have the correct time, but the internal `current_time` is a step back, so this assignment caused the difference.

In order to solve this problem, I've added a step increment to the internal `current_time_` value when the `EntityManager` update is ended. I also considered removing the override in `getEntityStatus` - but I'm not sure why it was created.
https://github.com/tier4/scenario_simulator_v2/blob/1770cbdbbfefc8216abed9ecc1622293d4b0e720/simulation/traffic_simulator/src/entity/entity_manager.cpp#L682

--

Unfortunately, despite obtaining identical steps of the algorithm - changes in time, distance, speed and acceleration are identical. This problem with the exception was still occurring.

The problem was related to the situation where the `remaining_time` variable was assigned a value of `0` at a very small `distance` from the goal - on the order of `1e-5` (so still not epsilon), so not small enough to skip the calculation of the `remaining_time` in this step.

Therefore, because of the `distance` that is still left, another step of the algorithm is performed in which `remaining_time` reaches a **value** `0` - **for the developed PR**. In the **master it reaches a value of the order of `1e-14`** (practically zero) - logs below.
For this reason, further calculations on the master do not throw an exception (there is no division by zero), and in the developed version exception occurs. 

The solution I've developed is to add `epsilon()` to the `remaining_time` variable when it reaches a value of `0`, which can cause an exception. This will not change the operation of the algorithm, but it will avoid the problem that was described. 
https://github.com/tier4/scenario_simulator_v2/blob/1770cbdbbfefc8216abed9ecc1622293d4b0e720/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L205-L208

#### Logs
The logs refer to the values of the arguments with which the function was called and the values that in a given step have variables in line:
https://github.com/tier4/scenario_simulator_v2/blob/1770cbdbbfefc8216abed9ecc1622293d4b0e720/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L288

**Master**
Screenshot from 2023-09-01 14-52-26
**Without fix**
Screenshot from 2023-09-01 15-03-04
**Fixed**
Screenshot from 2023-09-01 14-51-57

I ran the regression test again for the current version.

hakuturu583 (approved):
Great work!
It looks good to me.

PR #1075, opened 2023-08-18
feat(openscenario_interpreter): Detail the message in case of scenario failure

dmoszynski:
## Types of PR

- [x] New Features
- [ ] Bugfix


openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -185,85 +185,86 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
-        * ------------------------------------------------------------------- */
+                       *
+                       *  This is the default initialization state of a StoryboardElement.
+                       *  When it is in this state, the runtime instantiation of the
+                       *  StoryboardElement is ready to execute once given a startTrigger. A
+                       *  runtime instantiation of any StoryboardElement is created once its
+                       *  parent element is in the standbyState. From the standbyState, the
+                       *  Story element instantaneously transitions into the runningState.
+                       *
+                       * ------------------------------------------------------------------- */
```
yamacir-kit, line 211:
```suggestion
        *
        *  This is the default initialization state of a StoryboardElement.
        *  When it is in this state, the runtime instantiation of the
        *  StoryboardElement is ready to execute once given a startTrigger. A
        *  runtime instantiation of any StoryboardElement is created once its
        *  parent element is in the standbyState. From the standbyState, the
        *  Story element instantaneously transitions into the runningState.
        *
        * ------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -185,85 +185,86 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
-        *  instantiation is now starting. The startTransition can be used in
-        *  conditions to trigger based on this transition.
-        *
-        * ------------------------------------------------------------------- */
+                          *
+                          *  The startTransition symbolizes that the execution of the runtime
+                          *  instantiation is now starting. The startTransition can be used in
+                          *  conditions to trigger based on this transition.
+                          *
+                          * ------------------------------------------------------------------- */
```
yamacir-kit, line 224:
```suggestion
        *
        *  The startTransition symbolizes that the execution of the runtime
        *  instantiation is now starting. The startTransition can be used in
        *  conditions to trigger based on this transition.
        *
        * ------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -185,85 +185,86 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
+                       *
+                       *  Act
+                       *    An Act's goal is accomplished when all its ManeuverGroups are in
+                       *    the completeState.
+                       *
+                       *  Story
+                       *    A Story's goal is accomplished when all its Acts are in the
+                       *    completeState.
+                       *
+                       * ------------------------------------------------------------------- */
```
yamacir-kit, line 267:
```suggestion
        *
        *  The runningState symbolizes that the execution of the runtime
        *  instantiation is now ongoing and has not yet accomplished its goal.
        *
        *  The concept of accomplishing a goal varies depending on the type of
        *  StoryboardElement under consideration:
        *
        *  Action
        *    An Action's goal is a function of the Action type and cannot be
        *    generalized. Accomplishing an Action's goal will involve meeting
        *    some arbitrary prerequisites related with the Action type (for
        *    example, a SpeedAction accomplishes its goal when the considered
        *    Entity is travelling at the prescribed speed). If an Action is
        *    acting on an EntitySelection, all instances of Entity within the
        *    selection have to complete in order to reach the completeState of
        *    the Action.
        *
        *  Event
        *    An Event's goal is accomplished when all its Actions are in the
        *    completeState.
        *
        *  Maneuver
        *    A Maneuver's goal is accomplished when all its Events are in the
        *    completeState.
        *
        *  ManeuverGroup
        *    A ManeuverGroup's goal is accomplished when all its Maneuvers are
        *    in the completeState.
        *
        *  Act
        *    An Act's goal is accomplished when all its ManeuverGroups are in
        *    the completeState.
        *
        *  Story
        *    A Story's goal is accomplished when all its Acts are in the
        *    completeState.
        *
        * ------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -272,67 +273,67 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
-        * -------------------------------------------------------------------- */
+                        *
+                        *  The endTransition occurs when the runtime instantiation of the
+                        *  StoryboardElement accomplishes its goal. Once the endTransition
+                        *  occurs, a check for completeness is made. A positive outcome moves
+                        *  the state machine to the completeState, whereas a negative outcome
+                        *  moves the state machine to the standbyState. The endTransition can
+                        *  be used in conditions to trigger based on this transition.
+                        *
+                        * -------------------------------------------------------------------- */
```
yamacir-kit, line 284:
```suggestion
        *
        *  The endTransition occurs when the runtime instantiation of the
        *  StoryboardElement accomplishes its goal. Once the endTransition
        *  occurs, a check for completeness is made. A positive outcome moves
        *  the state machine to the completeState, whereas a negative outcome
        *  moves the state machine to the standbyState. The endTransition can
        *  be used in conditions to trigger based on this transition.
        *
        * -------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -272,67 +273,67 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
+                        *  upon finishing the runningState. This check returns false if there
+                        *  are executions left. This check returns true if there are no
+                        *  executions left, or if the maximumExecutionCount is not defined in
+                        *  the StoryboardElement.
+                        *
+                        *  Resetting the completeState can only be achieved externally by the
+                        *  parent StoryboardElement whose child is in the completeState. This
+                        *  may only occur if the parent initiates a new execution.
+                        *
+                        * -------------------------------------------------------------------- */
```
yamacir-kit, line 309:
```suggestion
        *
        *  The completeState signals that the runtime instantiation of the
        *  StoryboardElement cannot reach a running state without external
        *  interference. If the affected runtime instantiation of the
        *  StoryboardElement is defined with a maximumExecutionCount, to be
        *  complete implies that there are no more executions left to run, or a
        *  stopTransition has occurred.
        *
        *  Checking for completeness involves verifying if the given runtime
        *  instantiation of the StoryboardElement still has executions left
        *  upon finishing the runningState. This check returns false if there
        *  are executions left. This check returns true if there are no
        *  executions left, or if the maximumExecutionCount is not defined in
        *  the StoryboardElement.
        *
        *  Resetting the completeState can only be achieved externally by the
        *  parent StoryboardElement whose child is in the completeState. This
        *  may only occur if the parent initiates a new execution.
        *
        * -------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -272,67 +273,67 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
-        *  used in conditions to trigger based on this transition.
-        *
-        * ------------------------------------------------------------------- */
+                         *
+                         *  Transition marking the moment an element is asked to move to the
+                         *  runningState but is instead skipped so it remains in the
+                         *  standbyState (only for Event instances). The skipTransition can be
+                         *  used in conditions to trigger based on this transition.
+                         *
+                         * ------------------------------------------------------------------- */
```
yamacir-kit, line 319:
```suggestion
        *
        *  Transition marking the moment an element is asked to move to the
        *  runningState but is instead skipped so it remains in the
        *  standbyState (only for Event instances). The skipTransition can be
        *  used in conditions to trigger based on this transition.
        *
        * ------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/storyboard_element.hpp
```diff
@@ -272,67 +273,67 @@ class StoryboardElement : private SimulatorCore::ConditionEvaluation
...
+                         *  This implies that the stopTransition cannot be reached other than
+                         *  with an external intervention to the runtime instantiation of the
+                         *  StoryboardElement.
+                         *
+                         *  When a runtime instantiation of a StoryboardElement goes through a
+                         *  stopTransition, all of its child elements are also forced to go
+                         *  through the same transition. The stopTransition can be used in
+                         *  conditions to trigger based on this transition.
+                         *
+                         * ------------------------------------------------------------------- */
```
yamacir-kit, line 336:
```suggestion
        *
        *  The stopTransition marks the reception of a stopTrigger or the
        *  storyboard element is overridden (applicable for Event and Action).
        *  This implies that the stopTransition cannot be reached other than
        *  with an external intervention to the runtime instantiation of the
        *  StoryboardElement.
        *
        *  When a runtime instantiation of a StoryboardElement goes through a
        *  stopTransition, all of its child elements are also forced to go
        *  through the same transition. The stopTransition can be used in
        *  conditions to trigger based on this transition.
        *
        * ------------------------------------------------------------------- */
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/trigger.hpp
```diff
@@ -28,7 +28,8 @@ inline namespace syntax
  *
  *  <xsd:complexType name="Trigger">
  *    <xsd:sequence>
- *      <xsd:element name="ConditionGroup" type="ConditionGroup" minOccurs="0" maxOccurs="unbounded"/>
+ *      <xsd:element name="ConditionGroup" type="ConditionGroup" minOccurs="0"
+ * maxOccurs="unbounded"/>
```
yamacir-kit, line 32:
```suggestion
 * <xsd:element name="ConditionGroup" type="ConditionGroup" minOccurs="0" maxOccurs="unbounded"/>
```

openscenario/openscenario_interpreter/src/syntax/act.cpp
```diff
@@ -31,6 +32,19 @@ Act::Act(const pugi::xml_node & node, Scope & scope)
   });
 }
 
+auto Act::run() -> void
+{
+  for (size_t index{0}; auto && maneuver_group : elements) {
```
yamacir-kit, line 37: This repository is compiled with C++17. Don't take advantage of C++20 features yet.

```suggestion
  std::size_t index = 0;
  for (auto && maneuver_group : elements) {
```

openscenario/openscenario_interpreter/src/syntax/init.cpp
```diff
@@ -28,8 +29,12 @@ auto Init::endsImmediately() const -> bool { return actions.endsImmediately(); }
...
 {
-  actions.startInstantaneousActions();
-  actions.runInstantaneousActions();
+  try {
+    actions.startInstantaneousActions();
+    actions.runInstantaneousActions();
+  } catch (const SpecialAction<EXIT_FAILURE> & action) {
+    throw SpecialAction<EXIT_FAILURE>("OpenSCENARIO.Storyboard.Init", action);
+  }
   return unspecified;
```
yamacir-kit, line 38:
```suggestion
  try {
    actions.startInstantaneousActions();
    actions.runInstantaneousActions();
    return unspecified;
  } catch (const SpecialAction<EXIT_FAILURE> & action) {
    throw SpecialAction<EXIT_FAILURE>("OpenSCENARIO.Storyboard.Init", action);
  }
```

openscenario/openscenario_interpreter/src/syntax/init_actions.cpp
```diff
@@ -87,20 +91,28 @@ auto InitActions::startInstantaneousActions() -> void
   for (auto && e : global_actions) {
     e.as<GlobalAction>().start();
   }
-  for (auto && e : user_defined_actions) {
-    auto & user_defined_action = e.as<UserDefinedAction>();
-    if (user_defined_action.endsImmediately()) {
-      user_defined_action.start();
+
+  for (size_t index{0}; auto && e : user_defined_actions) {
```
yamacir-kit, line 95: This repository is compiled with C++17. Don't take advantage of C++20 features yet.
```suggestion
    std::size_t index = 0;
    for (auto && user_defined_action : user_defined_actions) {
```

openscenario/openscenario_interpreter/src/syntax/maneuver.cpp
```diff
@@ -29,6 +30,19 @@ Maneuver::Maneuver(const pugi::xml_node & node, Scope & scope)
   });
 }
 
+auto Maneuver::run() -> void
+{
+  for (size_t index{0}; auto && event : elements) {
```
yamacir-kit, line 35: This repository is compiled with C++17. Don't take advantage of C++20 features yet.
```suggestion
    std::size_t index = 0;
    for (auto && event : elements) {
```

openscenario/openscenario_interpreter/src/syntax/maneuver_group.cpp
```diff
@@ -35,6 +36,19 @@ ManeuverGroup::ManeuverGroup(const pugi::xml_node & node, Scope & scope)
   });
 }
 
+auto ManeuverGroup::run() -> void
+{
+  for (size_t index{0}; auto && maneuver : elements) {
```
yamacir-kit, line 41: This repository is compiled with C++17. Don't take advantage of C++20 features yet.
```suggestion
  std::size_t index = 0;
  for (auto && maneuver : elements) {
```

openscenario/openscenario_interpreter/src/syntax/maneuver_group.cpp
```diff
@@ -35,6 +36,19 @@ ManeuverGroup::ManeuverGroup(const pugi::xml_node & node, Scope & scope)
...
+    try {
+      assert(maneuver.is_also<Maneuver>());
+      maneuver.evaluate();
+      index++;
```
yamacir-kit, line 45:
```suggestion
      ++index;
```

openscenario/openscenario_interpreter/src/syntax/story.cpp
```diff
@@ -34,6 +35,19 @@ Story::Story(const pugi::xml_node & node, Scope & scope)
   });
 }
 
+auto Story::run() -> void
+{
+  for (size_t index{0}; auto && act : elements) {
```
yamacir-kit, line 40: This repository is compiled with C++17. Don't take advantage of C++20 features yet.
```suggestion
  std::size_t index = 0;
  for (auto && act : elements) {
```

openscenario/openscenario_interpreter/src/syntax/story.cpp
```diff
@@ -34,6 +35,19 @@ Story::Story(const pugi::xml_node & node, Scope & scope)
...
+    try {
+      assert(act.is_also<Act>());
+      act.evaluate();
+      index++;
```
yamacir-kit, line 44:
```suggestion
      ++index;
```

openscenario/openscenario_interpreter/src/syntax/storyboard.cpp
```diff
@@ -44,8 +45,13 @@ Storyboard::Storyboard(const pugi::xml_node & node, Scope & scope)
 
 auto Storyboard::run() -> void
 {
-  for (auto && story : elements) {
-    story.evaluate();
+  for (size_t index{0}; auto && story : elements) {
```
yamacir-kit, line 48: This repository is compiled with C++17. Don't take advantage of C++20 features yet.
```suggestion
  std::size_t index = 0;
  for (auto && story : elements) {
```

openscenario/openscenario_interpreter/src/syntax/storyboard.cpp
```diff
@@ -44,8 +45,13 @@ Storyboard::Storyboard(const pugi::xml_node & node, Scope & scope)
...
+    try {
+      story.evaluate();
+      if (story.is<Story>()) index++;
```
yamacir-kit, line 51:
```suggestion
      if (story.is<Story>()) {
        ++index;
      }
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/trigger.hpp
```diff
@@ -47,6 +48,9 @@ struct Trigger : public std::list<ConditionGroup>
   {
   }
 
+  auto activeConditionGroupIndex() const -> int;
```
yamacir-kit, line 51:
```suggestion
  auto activeConditionGroupIndex() const -> iterator::difference_type;

```

openscenario/openscenario_interpreter/src/syntax/trigger.cpp
```diff
@@ -44,6 +44,30 @@ auto Trigger::evaluate() -> Object
...
+{
+  size_t index{0};
+  for (auto it = begin(); it != end(); ++it, ++index) {
+    ConditionGroup const & cgroup = *it;
+    bool all_conditions_true = std::all_of(
+      cgroup.begin(), cgroup.end(), [](const Condition & c) { return c.current_value; });
+    if (all_conditions_true) break;
+  }
+  return index;
+}
```
yamacir-kit, line 57:
```suggestion
auto Trigger::activeConditionGroupIndex() const -> iterator::difference_type
{
  return std::distance(begin(), std::find_if(begin(), end(), [](const auto & condition_group) {
                         return std::all_of(
                           condition_group.begin(), condition_group.end(),
                           [](const auto & condition) { return condition.current_value; });
                       }));
}
```

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/custom_command_action.hpp
```diff
@@ -32,6 +32,195 @@ struct SpecialAction : public std::integral_constant<int, Value>
...
+    return ss.str();
+  }
+
+  const std::shared_ptr<const SpecialAction> inner_{nullptr};
+  const std::shared_ptr<const CoreSource> core_{nullptr};
+  const std::string source_name_{""};
+  const std::string path_{""};
+  const std::string log_{"No core source - the path is empty"};
+};
+
```
yamacir-kit, line 223: The class works fine, but is overly complex for the simplicity of its role. I was able to refactor the code to make it somewhat better. However, I am still not satisfied.

The problem with this class is that it is hard to imagine from the code the behavior of a gradual expansion of error messages from the depths of the scenario tree to the top level. In particular, each constructor corresponds to one or more specific elements of OpenSCENARIO, but it is hard to tell.

Better code is expected to result from modifying the constructors to accept pointers to element types.
In this case, each element would pass `this` pointer to the `SpecialAction` argument to construct the `SpecialAction` and throw or rethrow it.
In other words, it is necessary to express a clear correspondence between the element type and the respective constructor of SpecialAction.

(Pointers to element types are `Act*`, `ConditionGroup*`, etc. Perhaps a forward declaration is needed.)

However, we do not have enough time to spend on developing this feature, so such modifications will be a future issue.

```suggestion
template <>
class SpecialAction<EXIT_FAILURE> : public std::integral_constant<int, EXIT_FAILURE>,
                                    public std::runtime_error
{
  const std::function<std::string(const std::string &)> make_message;

  const std::string source_name;

  const std::string path;

  template <typename... Ts>
  static auto concatenate(Ts &&... xs)
  {
    std::stringstream ss;
    (ss << ... << xs);
    return ss.str();
  }

  template <typename Container>
  struct comma_separated
  {
    const Container & container;

    const std::function<void(std::ostream &, const typename Container::value_type &)> print;

    template <typename Printer>
    explicit comma_separated(const Container & container, Printer && print)
    : container{container}, print{std::forward<decltype(print)>(print)}
    {
    }

    friend auto operator<<(std::ostream & os, const comma_separated & cs) -> std::ostream &
    {
      auto separtor = "";
      for (const auto & value : cs.container) {
        os << std::exchange(separtor, ", ");
        cs.print(os, value);
      }
      return os;
    }
  };

  template <typename Conditions>
  static auto makeMessage(
    const std::string & trigger_name, const std::string & trigger_path,
    const Conditions & conditions)
  {
    std::stringstream what;

    what << "CustomCommandAction typed " << std::quoted("exitFailure") << " was triggered by the ";

    const auto anonymous = std::any_of(
      conditions.begin(), conditions.end(),
      [](const auto & condition) { return condition.first.empty(); });

    switch (conditions.size()) {
      case 0:
        what << trigger_name << " (" << trigger_path << ")";
        break;

      case 1:
        if (anonymous) {
          what << "anonymous " << trigger_name << " (" << trigger_path << "." << trigger_name
               << "[0]): " << conditions[0].second;
        } else {
          what << trigger_name << " named " << std::quoted(conditions[0].first) << ": "
               << conditions[0].second;
        }
        break;

      default:
        if (anonymous) {
          what << "anonymous " << trigger_name << "s (" << trigger_path << "." << trigger_name
               << "[...]): "
               << comma_separated(conditions, [i = 0](auto & os, const auto & condition) mutable {
                    os << '{' << std::quoted(std::to_string(i++)) << ": " << condition.second
                       << '}';
                  });
        } else {
          what << "named " << trigger_name << "s {"
               << comma_separated(
                    conditions,
                    [](auto & os, const auto & condition) { os << std::quoted(condition.first); })
               << "}: " << comma_separated(conditions, [](auto & os, const auto & condition) {
                    os << '{' << std::quoted(condition.first) << ": " << condition.second << '}';
                  });
        }
        break;
    }

    return what.str();
  }

public:
  SpecialAction()
  : std::runtime_error{
      "CustomCommandAction typed \"exitFailure\" was triggered on an unexpected code path. This is "
      "a simulator bug. Report this to the simulator developer."}
  {
  }

  explicit SpecialAction(
    const std::function<std::string(const std::string &)> & make_message,
    const std::string & source_name, const std::string & path)
  : std::runtime_error{make_message(path)},
    make_message{make_message},
    source_name{source_name},
    path{path}
  {
  }

  // Constructor with no inner object - first object with conditions
  explicit SpecialAction(
    const std::string & source_name, const std::string & element_name, int element_index,
    const std::string & trigger_name,
    const std::vector<std::pair<std::string, std::string>> & conditions = {})
  : SpecialAction{
      [trigger_name, conditions](const auto & trigger_path) {
        return makeMessage(trigger_name, trigger_path, conditions);
      },
      source_name, concatenate('.', element_name, '[', element_index, ']')}
  {
  }

  // Constructor with inner object and current source (to be forwarded) - intermediate object
  explicit SpecialAction(
    const std::string & source_name, const std::string & element_name, int element_index,
    SpecialAction const & inner)
  : SpecialAction{
      inner.make_message, source_name,
      concatenate(
        '.', element_name, '[',
        (inner.source_name.empty() ? std::to_string(element_index) : '"' + inner.source_name + '"'),
        ']', inner.path)}
  {
  }

  // Constructor with inner but without current source -- last object
  explicit SpecialAction(const std::string & element_name, const SpecialAction & inner)
  : SpecialAction{
      inner.make_message, "",
      concatenate(
        element_name, (inner.source_name.empty() ? "" : "." + inner.source_name), inner.path)}
  {
  }
};
```

yamacir-kit (changes requested):

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/custom_command_action.hpp
dmoszynski: @yamacir-kit 
I've introduced the refactored version of `SpecialAction` you suggested.
I ran local tests - the results of the scenarios are identical.
Perhaps it was indeed redundant to create `CoreSource` structure and its passing, but certainly the `inner_` object was redundant...

I admit that your approach is less complex, I didn't think of this approach, **thank you so much.**


> Better code is expected to result from modifying the constructors to accept pointers to element types.
> In this case, each element would pass this pointer to the SpecialAction argument to construct the SpecialAction and throw or rethrow it.
> In other words, it is necessary to express a clear correspondence between the element type and the respective constructor of SpecialAction.

This approach is also interesting and will indeed introduce more clarity. I will be glad to take care of it in the future.

Changes since the last regression tests are [in this commit](https://github.com/tier4/scenario_simulator_v2/pull/1075/commits/149c6401a9cbfc80afb32fea502cb089bdec88bb). Please let me know if you think re-running the evaluation is necessary. If not, is there anything else I can do for you within this PR?

yamacir-kit: > Please let me know if you think re-running the evaluation is necessary. If not, is there anything else I can do for you within this PR?

I think I need to do it again because the amount of code changes is large... but I ended up merging it due to an operational error.
I think it's probably okay, but I'll test it once just to be sure.

PR #1077, opened 2023-08-24
Fix/autoware shutdown

yamacir-kit:
## Types of PR

- [ ] New Features
- [x] Bugfix

## Description

Fixed two known issues with simulator shutdown of Autoware.

### 1

When the simulator was shutting down Autoware, if an Autoware node became unresponsive due to a segmentation fault or other reason, the simulator could wait indefinitely for that node. This pull request changes such a case to send a SIGKILL only once to the process group that includes the launch process of Autoware, and then proceed to the subsequent process. This change also eliminates the problem that Autoware nodes were sometimes not properly terminated (typically, the rviz2 process was often not terminated).

### 2

In the simulator shutdown process, there is a process that takes a very long time. During such a process, the simulator becomes unresponsive and is judged "not good" by `status_monitor`. This is believed to be the cause of the "health check failed" problem that has been reported recently on Autoware Evaluator. To address this, this pull request changes the threshold at which status_monitor judges "not good" when starting a long process (`common::StatusMonitor::overrideThreshold`).


yamacir-kit:
**TODO: Change the default value of sigterm_timeout to a number that allows a safe shutdown with Autoware's minimum machine specs.**

yamacir-kit:
**TODO: When promoting SIGINT to SIGKILL, kill the process group instead of the ros2 launch process.**

yamacir-kit:
> TODO: When promoting SIGINT to SIGKILL, kill the process group instead of the ros2 launch process.

Done.

PR #1099, opened 2023-10-03
feat(concealer, scenario_test_runner): ss2/AWSIM connection: initial pose sending upgrade, sample scenario

piotr-zyskowski-rai:
## Types of PR

- [ ] New Features
- [X] Upgrade of existing features
- [ ] Bugfix

## Link to the issue
https://tier4.atlassian.net/browse/RJD-108

## Description
This PR adds:
- feature to send initial pose to Autoware only when localization system is in an UNINITIALIZED state
- sample scenario to run with AWSIM

## How to review this PR.
Change is relatively small; review as a diff should be convenient.

## Others
It is the continuation of work related to connecting AWSIM with scenario_simulator_v2


external/concealer/include/concealer/field_operator_application_for_autoware_universe.hpp
```diff
@@ -15,6 +15,7 @@
 #ifndef CONCEALER__AUTOWARE_UNIVERSE_USER_HPP_
 #define CONCEALER__AUTOWARE_UNIVERSE_USER_HPP_
 
+#include <autoware_adapi_v1_msgs/msg/localization_initialization_state.hpp>
```
yamacir-kit, line 18: This message type exists only in certain newer versions of `autowarefoundation/autoware_adapi_msgs`. We may assume the existence of a message package such as `autoware_adapi_msgs` in our source tree, but we must not assume its existence for that individual message type. This is because detecting the existence of ROS 2 packages is hard at the C++ code or CMake script level, while the existence of message types can be easily detected at the C++ level by using `#if __has_include`. In other words, the policy is to provide as much backward compatibility as possible with reasonable effort.

Therefore, please enclose code that depends on the `autoware_adapi_v1_msgs::msg::LocalizationInitializationState` introduced by this pull request by `#if __has_include`.

yamacir-kit (changes requested):

external/concealer/include/concealer/field_operator_application_for_autoware_universe.hpp
piotr-zyskowski-rai: I added the flags :+1:

yamacir-kit: Since `scenario_simulator_v2` requires that it be compiled with standard C++17, there is no need to check whether `__has_include`, a language feature of the C++17 standard, is defined. (This is a problem in the case of `#undef __has_include`, but you can exclude the possibility of such terrible code written.)

Also (although this is a matter of preference) there is no need to use macros like `USE_AUTOWARE_LOCALIZATION_STATE` as an intermediary. I think it is clearer to use `#if __has_include(<autoware_adapi_v1_msgs/msg/localization_initialization_state.hpp>)` directly in each place. I think the person who sees this code of yours probably doesn't immediately know what `USE_AUTOWARE_LOCALIZATION_STATE` means (because the code is related to "the initialization state of localization", not the "state of localization". I think the person will end up checking the definition of `USE_AUTOWARE_LOCALIZATION_STATE` using grep etc.)

Note: I was surprised by your code for `#if defined __has_include` because I remembered that `__has_include` was an expression of a language primitive, not a macro. When I experimented with minimal code, I was able to surprisingly confirm that `#if defined __has_include` is true (in C++17 and later). I was curious, so I looked into the standard again and found that although `__has_include` is an expression and not a macro name, it is specially treated as if it were a macro name in the context of preprocessor directives such as `#if` (https://en.cppreference.com/w/cpp/preprocessor/conditional).

piotr-zyskowski-rai: Thank you for the rereview and a great explanation :bow: 

> Since scenario_simulator_v2 requires that it be compiled with standard C++17, there is no need to check whether __has_include, a language feature of the C++17 standard, is defined

Thank you for pointing that out. I missed that. I removed the `#if defined` clause

> Also (although this is a matter of preference) there is no need to use macros like USE_AUTOWARE_LOCALIZATION_STATE as an intermediary

I would like to share the rationale behind using this name. I used the intermediate macro to achieve two things:
- the macro allows to mention the specific header only in one place, which would make it easier to modify the header without the need to make sure, that all headers are modified within the code. 
- The name of the macro can carry additional information on why it was necessary. Of course, it has to be informative in the context. In hindsight, the name `USE_AUTOWARE_LOCALIZATION_STATE` might not convey the proper message. Something like `AUTOWARE_LOCALIZATION_STATE_INFORMATION_AVAILABLE` might reflect the message more clearly.

This rationale, however, is not strong enough for me to stick to the original idea, hence I removed the `USE_AUTOWARE_LOCALIZATION_STATE` macro to match your preference.

yamacir-kit (approved):

PR #1112, opened 2023-10-20
Fix/Standstill duration condition for MiscObjects

piotr-zyskowski-rai:
# Description

## Abstract

Added standstill duration update for MiscObjects. It is necessary to support scenarios that are testing the standstill duration for those objects.

## Background

Standstill duration of the misc object was not updated, so it is always 0.
It was a bug.

## Details

Update standstill duration of the misc object and fix the bug.

## References

https://github.com/tier4/sim_evaluation_tools/issues/214

[OpenSCENARIO documentation link.
](https://www.asam.net/static_downloads/ASAM_OpenSCENARIO_V1.2.0_Model_Documentation/modelDocumentation/content/StandStillCondition.html)
# Destructive Changes

Standstill duration of the MiscObject was always 0.
But after this pull request merged, it become greater than or equal to 0.

# Known Limitations

N/A


yamacir-kit:
I have verified that by applying the changes in this pull request, the workflow https://github.com/RobotecAI/scenario_simulator_v2_scenarios/blob/main/workflow/MiscObject.yaml now all succeed.

hakuturu583 (approved):

hakuturu583 (changes requested):
Code is LGTM.
But, did you tested it on Evaluator?

hakuturu583 (approved):

PR #1124, opened 2023-11-02
[WIP] Add a command to procedurally generate intersections

Goldob:


Goldob:
@yamacir-kit 

The PR has been rebased to account for changes in underyling library (#1108). The feature of generating X-shaped intersection has been completed and is ready for review. Usage of the newly added command follows below.

## `intersection.cpp`

The command generates an X-shaped intersection of given geometry and structure.

**Geometry** is defined by parameters :`leg_length` and `turn_radius`, as seen on the following diagram:

image

**Structure** means layout of lanes for each intersection leg: how many lanes are there and in which directions do they allow to leave the intersection (left/straight/right). In current implementation, the following simplifying assumptions were made:
- each intersection leg has the same amount and layout of lanes;
- for each intersection leg, there is the same amount of lanes for both driving directions;
- i-th lane of an inbound leg can only be connected to an i-th leg of any outbound leg.

Structure is defined by parameters:
- `number_of_lanes_going_left`
- `number_of_lanes_going_left_and_straight`
- `number_of_lanes_going_straight`
- `number_of_lanes_going_in_all_directions`
- `number_of_lanes_going_straight_and_right`
- `number_of_lanes_going_right`

The ordering of lanes (left-to-right) within an inbound leg follows the above sequence. However, `number_of_lanes_going_straight` cannot be used in conjunction with `number_of_lanes_going_left_and_straight`, `number_of_lanes_going_straight`, or `number_of_lanes_going_straight_and_right`, because it would lead to undefined behavior.

### Examples

```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_in_all_directions:=2
```

image

```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_left_and_straight:=1 \
-p number_of_lanes_going_straight:=1 \
-p number_of_lanes_going_right:=1
```

image

Goldob:
To implement connections between subsequent road segments (necessary for routing), a new struct called `RoadSegmentConnection` has been introduced. Lanelets generated from segments connected in this way will share end/start points. Code for achieving that was complicated enough to warrant moving it completely into a separate file (`generate_lanelet_map.hpp`). A nice side effect is that we now have a clear separation between the two abstractions (road segments and lanelets). You can think in the former and only convert to the latter at the very last step.

The growing number of header files also made me do a bit of reorganizing the directory structure. However, because of that, the diff of this PR can be difficult to read. If it makes the review to difficult to do, commit [48e70d6](https://github.com/tier4/scenario_simulator_v2/pull/1124/commits/48e70d65089e94f948cd168ba2f6d2a99ff3cd8f) can be reverted.

openscenario/map_fragment/src/intersection.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  const auto number_of_lanes_going_straight = [&]() {
+    node.declare_parameter("number_of_lanes_going_straight", 0);
+    return node.get_parameter("number_of_lanes_going_straight").as_int();
+  }();
+
+  const auto number_of_lanes_going_in_all_directions = [&]() {
+    node.declare_parameter("number_of_lanes_going_in_all_directions", 0);
+    return node.get_parameter("number_of_lanes_going_in_all_directions").as_int();
+  }();
```
yamacir-kit, line 77: If the value given to `number_of_lanes_going_in_all_directions` exceeds 6, a malformed intersection is generated (This threshold may vary depending on the value given to `turn_radius`.) This problem is unacceptable because 8-lane intersection actually exists. In addition, please make sure that the proper intersection is generated even if a value too large, such as 100, is given. This is because the inability to generate the proper intersection in such a case implies that there is a code somewhere that is making incorrect assumptions.

yamacir-kit (changes requested):
Do not set the default values of parameters to values that would cause the intersection to fail to be generated. It is desirable that a minimum configuration of intersections be generated when `intersection.cpp` is run with no arguments.

openscenario/map_fragment/src/intersection.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  const auto leg_length = [&]() {
+    node.declare_parameter("leg_length", default_value::length);
+    return node.get_parameter("leg_length").as_double();
+  }();
+
+  const auto turn_radius = [&]() {
+    node.declare_parameter("turn_radius", default_value::turn_radius);
+    return node.get_parameter("turn_radius").as_double();
+  }();
```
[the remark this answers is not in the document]
Goldob: [Done](https://github.com/tier4/scenario_simulator_v2/pull/1124/commits/6c9501174370384f258595e143868496cc9672be).

openscenario/map_fragment/src/intersection.cpp
Goldob: You are correct, the malformed intersection is due to too small value of `turn_radius` parameter. If [the newly added check](https://github.com/tier4/scenario_simulator_v2/pull/1124/files#diff-9f581456265172861eea619b8d2400cf8775f29cbbf37b15b9b6c5fa0fa54ccbR102-R109) fails, it will give hints what minimal value should be provided for given configuration.

I wasn't able to load a 100-lane intersection into Scenario Editor to validate it (or perhaps I was just too impatient to wait for so long), but here is a 50-lane intersection (minimal turn radius in this case is 500.0):

```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_in_all_directions:=50 \
-p turn_radius:=600.0
```

image

It looks to me like a microcontroller.

Goldob:
> Do not set the default values of parameters to values that would cause the intersection to fail to be generated. It is desirable that a minimum configuration of intersections be generated when intersection.cpp is run with no arguments.

I did the other way on purpose. This is because if any of the parameters is assigned a default non-zero value, it will have to be explicitly reset if any other structure is desired.

For example, if `number_of_lanes_going_in_all_directions` has a default value of, say, 2...
Instead of this:
```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_left_and_straight:=1 \
-p number_of_lanes_going_straight:=1 \
-p number_of_lanes_going_right:=1
```

We would need to do this:
```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_in_all_directions:=0 \
-p number_of_lanes_going_left_and_straight:=1 \
-p number_of_lanes_going_straight:=1 \
-p number_of_lanes_going_right:=1
```
I think the former is more clean.

yamacir-kit:
@Goldob 
In general, it is not good to have dependencies between options. Options should be in an orthogonal relationship. From the looks of it, it seems that the `number_of_lanes_going_in_all_directions` added for convenience is the cause. Would it be possible to eliminate the dependency between parameters by eliminating this option? (I don't mind if it reduces convenience. I believe that now is the phase where orthogonality is a better priority.)

Alternatively, would it solve the problem if `intersection.cpp` were implemented to detect if each parameter is given or not, as the code in `src/array.cpp` in the `feature/map-fragment` branch does?

Goldob:
@yamacir-kit 

No other option (or combination of options) can replace `number_of_lanes_going_in_all_directions`. This is the only way to specify lanes which can exit in all directions. All other lanes can exit only in two or a single direction.

The ambiguity I mentioned between `number_of_lanes_going_straight` and `number_of_lanes_going_in_all_directions` comes from the fact that it is not clear how these two lanes should be ordered w.r.t. one another. All other lanes have a natural, intuitive ordering. Starting from left to right we have:
- lanes which only go left (1 direction)
- lanes which go left and straight (2 directions)
- lanes which go straight (1 direction) **OR** lanes which go left, straight, and right (all 3 directions)
- lanes which go straight and right (2 directions)
- lanes which go right (1 direction)

Goldob:
To illustrate the difference:

```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_left:=1 \
-p number_of_lanes_going_straight:=1 \
-p number_of_lanes_going_right:=1
```

image

```
ros2 run map_fragment intersection --ros-args \
-p number_of_lanes_going_in_all_directions:=3
```

image

yamacir-kit:
@Goldob 

I understood that the problem stems from the complexity of the intersection itself. Let's make this an issue for the next term. As we add new features to `intersection.cpp`, we may be able to see what it should be.

However, if a parameter is given that is inconsistent or inconvenient, please be sure to notify us of this as an error with the reason for the error. It is acceptable if the problem appears as an "input error. It is unacceptable for a malformed map to be output when it appears at first glance that the input was accepted into the program without problems.

Goldob:
I agree, it was my oversight. In addition to [the check for `lane_width`](https://github.com/tier4/scenario_simulator_v2/pull/1124/commits/6c9501174370384f258595e143868496cc9672be), [I added two more checks to validate user input](https://github.com/tier4/scenario_simulator_v2/pull/1124/commits/ad8be8311c37e85ee152c23baa9f57ac988515c4).

openscenario/map_fragment/src/intersection.cpp
```diff
@@ -0,0 +1,277 @@
...
+  const auto half_leg_width = total_number_of_lanes_per_leg_per_direction * lane_width;
+
+  rcpputils::require_true(
+    turn_radius >= half_leg_width,
+    "Turn radius must not be smaller than half leg width. Actual values:\n"
+    "- Turn radius: " +
+      std::to_string(turn_radius) +
+      "\n"
+      "- Half leg width: " +
+      std::to_string(half_leg_width));
```
yamacir-kit, line 120: This is my rule of thumb, but error messages for programs that are not certain to be executed interactively on the terminal should not contain newline characters. For instance, this program is likely to run on Autoware Evaluator in the future, and Autoware Evaluator's error display UI assumes that error messages do not contain line breaks.

openscenario/map_fragment/src/intersection.cpp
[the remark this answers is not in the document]
yamacir-kit: I tested the following commands and confirmed that an error occurs.

Command:
```
ros2 run map_fragment intersection --ros-args \
  -p number_of_lanes_going_in_all_directions:=2 \
  -p turn_radius:=10.0
```

Output:
```
Turn radius must not be smaller than half leg width. Actual values:
- Turn radius: 10.000000
- Half leg width: 20.000000
```

However, there is a problem in that the error message contains the term "half leg width," a term that one would not know what it means unless one knows the internal implementation. **I don't require this to be resolved in this pull request**, but when writing error messages, please do not use information that is not available to the user. This may be perceived by the user as a failure of an internal assertion rather than a failure based on the parameters given by the user.

yamacir-kit (approved):
Although the amount of code is large, I did not find the code to be redundant, as the comments and assertions take up a large percentage of the code. However, I think it is likely to induce programming mistakes due to the large number of special processes based on the number of lanes. For example, even you, the original author of this code, would probably have a hard time grasping the intent of the code if you tried to revise this code after a period of several months.

In other words, the internal representation of the intersection needs to be refined. However, it is premature to do that in this pull request. This is because the intersections generated by this command still need some additional functionality to be practical. For example, the attribute `turn_direction` needs to be set for the straight, left-turn, and right-turn lanes of the intersection, respectively. We also need an option to relate traffic signals to those lanes. Refining the internal representation of intersections should be done after those are implemented. The same goes for organizing the parameters of the commands.

Therefore, there is no additional work now, so I approve this pull request.

PR #1128, opened 2023-11-08
Random test runner fix

pawellech1:
## Types of PR

- [ ] New Features
- [X] Bugfix

## Link to the issue
https://tier4.atlassian.net/browse/RJD-706

## Description
Multiple problems were fixed in this PR in order to allow performing tests with the random test runner successfully:

- The duration of Autoware initialization was not specified in the random test runner launch file. Because of that test runner was not waiting for Autoware to initialize and instead stopped immediately. The default value of the parameter is now set inside the launch file.
- `allow_goal_modification` was not declared as a parameter which made concealer crash while it was trying to use the parameter. The parameter is now declared with the usage of `traffic_simulator::API`. 
- `OccupancyGridSensor` was not attached by `TestExecutor` which blocked Autoware in planning the route. Autoware was stuck waiting for the initialization of occupancy grid map. The sensor is now attached with the usage of using `traffic_simulator::API`.
- Because all `TestExecutors` were using the same instance of `traffic_simulator::API` only the first test could be initialized properly. It was caused by the fact that `EntityManager` does not allow to set the entity status after starting the scenario. Separate `traffic_simulator::API` instances are now being attached to each `TestExecutor`.
- Exception thrown during the test ended the random test runner execution. Exceptions handling in `TestExecutor` is now added so that an exception ends only currently executed test and makes random test runner proceed to the next test.

## Others
There is an issue which blocks random test runner from initializing the test if either initial position or goal pose exceeds the drivable area. 
- In the first case Autoware will be launched but the ego vehicle will not move waiting for Autoware to change its state to `WaitingForEngage`. After a timeout the test will finish, the error will be logged and the runner will continue with the next test.
- In the second case test will be immediately stopped after launching the Autoware and similarly as in the first case the error will be logged and the runner will continue with the next test.

This issue is not related with this PR.

### DIFF
--- a/test_runner/random_test_runner/include/random_test_runner/file_interactions/junit_xml_reporter.hpp
+++ b/test_runner/random_test_runner/include/random_test_runner/file_interactions/junit_xml_reporter.hpp
@@ -40,6 +40,11 @@ public:
 
   void reportTimeout() { reportError("timeout", "Ego failed to reach goal within timeout"); }
 
+  void reportException(const std::string & type, const std::string & error_msg)
+  {
+    reportError(type, error_msg);
+  }
+
 private:
   void reportError(const std::string & error_type, const std::string & message)
   {
--- a/test_runner/random_test_runner/include/random_test_runner/test_executor.hpp
+++ b/test_runner/random_test_runner/include/random_test_runner/test_executor.hpp
@@ -41,6 +41,8 @@ public:
   bool scenarioCompleted();
 
 private:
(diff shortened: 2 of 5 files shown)


hakuturu583 (changes requested):
Did you run regression test in Autoware Evaluator?

hakuturu583:
Sorry, I found that it is unnecessary for checking regression test in Evaluator.

hakuturu583 (approved):

hakuturu583:
LGTM!

Issue #1123, opened 2023-10-31
CatmullRomSubspline collision not being detected


TauTheLepton:
**Describe the bug**
The `CatmullRomSubspline` member function `getCollisionPointIn2D` does not always return the correct result. In a simple case with one collision is works fine, but in more complicated scenarios with more collision points the function does not always return a collision point.
This is caused by the fact, that the `CatmullRomSubspline` calls the `CatmullRomSpline` member function `getCollisionPointIn2D` here:
https://github.com/tier4/scenario_simulator_v2/blob/c4b7daa714f2d7b94d6a71c0f78932b0826b4ca3/common/math/geometry/src/spline/catmull_rom_subspline.cpp#L28
which already filters out one collision point here (the first one found is returned):
https://github.com/tier4/scenario_simulator_v2/blob/c4b7daa714f2d7b94d6a71c0f78932b0826b4ca3/common/math/geometry/src/spline/catmull_rom_spline.cpp#L298-L320
This leads to incorrect situations, where a polygon has multiple collisions with the spline, the spline returns first (or last - depending on the `search_backward` argument) collision point, but this collision point is not part of the subspline. Some other collision point may be a part of the subspline, but this case is not checked, the information is lost and no collision point is returned from the `CatmullRomSubspline` member function `getCollisionPointIn2D`.

In my opinion there should exist a `CatmullRomSpline` member function that will return all collision points, so that from those collision points `CatmullRomSubspline` could choose the ones that are part of the spline and then return the first one (or the last one).
**NOTE** that this fix will require to modify the code that is affected by another bug (https://github.com/tier4/scenario_simulator_v2/issues/1120) which should probably be fixed first.

_Context:_
I am adding unit tests for the `geometry` package and wanted to add tests for the `getCollisionPointIn2D` function, but it generates the wrong results in some cases (see **To Reproduce**).

**To Reproduce**
I have created an example in the test case to demonstrate the bug [here](https://github.com/TauTheLepton/scenario_simulator_v2/blob/a781708f099297cbdb0a600cd14bb1eb3210ea2c/common/math/geometry/test/test_catmull_rom_spline.cpp#L357-L396)
1. Clone [this branch](https://github.com/TauTheLepton/scenario_simulator_v2/tree/example/subspline-collision-bug)
2. Build
3. Run tests
4. See that the last spline test passes (which makes no sens, as the same spline with the same polygon does collide in one case and does not collide in other case)

**Expected behavior**
[...]

Issue #1120, opened 2023-10-24
CatmullRomSpline incorrect collision point calculation


TauTheLepton:
**Describe the bug**
The `CatmullRomSpline` member function `getCollisionPointIn2D` calculates wrong values.
This member function uses its component curves member function `getCollisionPointIn2D` here:
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/catmull_rom_spline.cpp#L383
and here:
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/catmull_rom_spline.cpp#L391
and simply adds the obtained value to the sum of the preceding curves. This produces an incorrect result, because `HermiteCurve` operates on normalized lengths, meaning the s value is from range <0; 1> like in the function below.
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/hermite_curve.cpp#L122-L181
And `CatmullRomSpline` operates on absolute lengths, so not normalized ones. This means that the normalized value is added to not normalized ones - which makes the incorrect result.
Many member functions in the `HermiteCurve` class offer a `denormalize_s` argument which - when set to true - converts normalized s to the absolute value of s, independent from the curvature length.
In my opinion, the best solution would be to call the `HermiteCurve` member function `getCollisionPointIn2D` with such argument `denormalize_s` set to true. However this member function does not offer such a functionality, so I suggest adding it the same way it is implemented here:
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/hermite_curve.cpp#L194-L197

_Context:_
I am adding unit tests for the geometry package and wanted to add tests for the `getCollisionPointIn2D` function, but it generates the wrong results.

**To Reproduce**
1. Edit any existing source code file or create a new one
2. Create a `CatmullRomSpline` from points spaced apart for more than 2 units (like (0, 0), (2, 0) and (4, 0) )
3. Call the `getCollisionPointIn2D` member function with two points creating a parallel line to the spline (e.g. (0.1, 1) and (0.1, -1)
4. Print the result to the console
5. Compile and execute
6. See the incorrect result

**Expected behavior**
[...]

Issue #1119, opened 2023-10-24
getMaximum2DCurvature wrong curvature calculation


TauTheLepton:
**Describe the bug**
`CatmullRomSpline` member function `getMaximum2DCurvature` calculates the wrong value.
The member function finds the greatest max 2D curvature value from all curves that are included in the spline. This is done in the line below.
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/catmull_rom_spline.cpp#L570
However the max 2D curvature can be a positive as well as a negative value. The curvature magnitude is determined by the absolute value of the curvature value, which is considered in the function below.
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/hermite_curve.cpp#L303-L310
This means that the `CatmullRomSpline` member function `getMaximum2DCurvature` having two values: -10 and 2 will choose 2, however the value -10 corresponds to a much greater curvature, just in the other direction. For a spline having all component curves bending in one way (when all component max 2D curvatures are negative) this member function actually does the opposite of what is should - it selects the smallest curvature.
The logic present in `HermiteCurve` member function should be applied to a `CatmullRomSpline` member function as well.

_Context:_
I am adding unit tests for the `geometry` package and wanted to add tests for the `getMaximum2DCurvature` function, but it generates the wrong result in a case similar to the example above.

**To Reproduce**
1. Edit any existing source code file or create a new one
2. Create a `CatmullRomSpline` passing through points (0, 0), (0.5, 0.5), (1, 0), (2, -1) and (3, 0)
3. Call the `getMaximum2DCurvature` member function
4. Print and see the incorrect result

**Expected behavior**
Greatest **absolute** value should be selected instead of the greatest one.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]: Ubuntu
 - Browser [e.g. chrome, safari]: Firefox
 - Version [e.g. 22]: 22
 - ROS 2 version: Humble
 - DDS: CycloneDDS

hakuturu583:
Thanks, it is a bug.

hakuturu583:
#1139 was merged

Issue #1118, opened 2023-10-24
HermiteCurve getTrajectory wrong number of points


TauTheLepton:
**Describe the bug**
The `HermiteCurve` member function `getTrajectory` taking number of points as an argument generates a trajectory with one point more than requested. The cause is that the condition in the loop below is `<=` and should be `<`.
https://github.com/tier4/scenario_simulator_v2/blob/173a8c1d5e16b6e594de58d78cdb81a1d518ad82/common/math/geometry/src/spline/hermite_curve.cpp#L226
Changing the loop condition will require the logic to be changed, including changing the `t` calculation, so going from
```cpp
double t = static_cast<double>(i) / static_cast<double>(num_points);
```
to
```cpp
double t = static_cast<double>(i) / (static_cast<double>(num_points) - 1.0);
```

_Context:_
I am adding unit tests for the `geometry` package and wanted to add tests for the `getTrajectory` function, but it generates the wrong result.

**To Reproduce**
1. Edit any existing source code file or add a new one
2. Call `HermiteCurve` member function `getTrajectory` with some number of points.
3. Print to the console how many points are actually generated
4. See the difference

**Expected behavior**
The trajectory should have as many points as were requested.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]: Ubuntu
 - Browser [e.g. chrome, safari]: Firefox
 - Version [e.g. 22]: 22
 - ROS 2 version: Humble
 - DDS: CycloneDDS

hakuturu583:
Thanks, it is a bug.
Please fix it.

hakuturu583:
#1139 was merged

PR #1139, opened 2023-11-15
Fix/geometry bug fixes

TauTheLepton:
## Types of PR

- [ ] New Features
- [x] Bugfix

## Link to the issue
#1114 #1115 #1116 #1117 #1118 #1119 #1120 #1123 

## Description
I have fixed all bugs that were reported in the issues linked above.
- I have removed equal operators for `geometry_msgs::msg::Point` and `geometry_msgs::msg::Vector3`, because they were ambiguous.
- I have fixed the bug which caused the intersection functions using vector to look past the last element of the vector and return wrong results.
- I have fixed a bug where the `LineSegment` class could be constructed with a `geometry_msgs::msg::Vector3` of size = 0. This lead to initialization of `end_point` with `nan` values.
- I have fixed the `getMinValue` and `getMaxValue` bug where when an empty vector was passed the function tried to de-reference `vector.end()` and the whole program crashed.
- I have fixed a `getPolygon` bug which caused the function to generate incorrect number of points.
- I have added support for negative curvature values which were already supported by `HermiteCurve`. This incompatibility lead to errors.
- I have fixed spline collision implementation error which caused spline to add normalized length to absolute lengths which is incorrect.
- I have fixed `CatmullRomSubspline` collision detection by enabling the `HermiteCurve` and `CatmullRomSpline` to have multiple collisions detected and then choosing in the subspline the collisions that occur inside the subspline.

### Curve length computation change
This PR fixes how the length of the curve is computed

| Before fix | After fix| 
|--------------|-----------|
| Lengths of the first two Curves (A and B) are calculated correctly (because these are full Curve lengths), but the distance in the third Curve (C) is calculated as half of the normalized length of the Curve (0.5). This value is added and the result and distance to the collision in the spline is equal to 20.5 | Lengths of the first two Curves (A and B) are calculated correctly (because these are full Curve lengths) and the distance in the third Curve (C) is also calculated correctly because the collision distance is being denormalized (multiplied by the full length of the Curve) which is equal to 5 (0.5 * 10).This value is added and the result and distance to the collision in the spline is equal to 25 | 
| RJD-753_base  | RJD-753_eval |

It might slightly influence the behavior of the scenarios. 

### Influenced traffic_simulator/OpenSCENARIO features and how to fix them

| traffic_simulator/OpenSCENARIO features | Severity | Fix | 
|----------------------------------------------------------|-------------|--------------|
| FollowFrontEntityAction | low - distance kept to the front entity might be a bit larger which might influence scenarios which are relying on specific behavior in this action but it does not seem to be a lot of scenarios | Conditions which check this distance should be increased accordingly - trial and error seem to be the best way to find how much change the condition without changing the essence of the scenario|
|  Computing the distance to stop_line or other "obstacles" | low - it will increase this distance making it a bit safer but if the scenario conditions rely on this value it might cause issues. No scenario, however, was noticed to be influenced by that to the extent that caused them to fail | Conditions that check this distance should be increased accordingly - trial and error seem to be the best way to find how much change the condition without changing the essence of the scenario |
| Distance-based lane change | Medium - regression is known which was caused by the fix. Cut-in scenarios which are written in a rather strict way might be influenced by this change | Decreasing the distance on which the lane change should occur or change the speed of the vehicles taking part in the scenario - example in the section below | 

### Scenario fix example

This scenario usually passes without the fix but always fails after the fix is added. The scenario is based on [shinjuku_map](https://github.com/tier4/AWSIM/releases/download/v1.2.0/shinjuku_map.zip).

[scenario.yml.txt](https://github.com/tier4/scenario_simulator_v2/files/13707779/scenario.yml.txt)

An issue in this scenario is lane change action:

```
LaneChangeAction:
  LaneChangeActionDynamics:
    dynamicsDimension: distance
    value: 10 
    dynamicsShape: cubic
  LaneChangeTarget:
    AbsoluteTargetLane:
      value: '37'
```

To fix it:
- The speed of Ego vehicle can be decreased slightly
- The speed of NPC motorcycle can be increased slightly
- The lane change distance can be decreased slightly

The exact amount of the change of the values above is hard to estimate because it is very dependent on the specific scenario - like the vehicle position within the lanelet might influence how much the normalized lanelet length was different than not normalized length.

Here is the scenario that uses the third possible fix - decreasing the distance on which the lane change action takes place. One meter decrease makes the scenario pass again.

```
LaneChangeAction:
  LaneChangeActionDynamics:
    dynamicsDimension: distance
    value: 9 
    dynamicsShape: cubic
  LaneChangeTarget:
    AbsoluteTargetLane:
      value: '37'
```

Full fixed scenario
[scenario.yml.txt](https://github.com/tier4/scenario_simulator_v2/files/13707839/scenario.yml.txt)


hakuturu583 (changes requested):
Perfect pull-request, thanks.
But it is a breaking change, so need to new version release.
Temporarily change to request changes to prevent accidental operation.

hakuturu583 (approved):

PR #998, opened 2023-05-03
feat(traffic_simulator):  ss2/AWSIM connection: ego entity split

piotr-zyskowski-rai:
## Types of PR

- [ ] New Features
- [x] Upgrade of existing features
- [ ] Bugfix

## Link to the issue
https://tier4.atlassian.net/browse/AJD-660

## Description
The goals of this PR are:
- split functionality inside `traffic_simulator::entity::EgoEntity` into the part that will stay in `traffic_simulator` and the part which will be moved to the `simple_sensor_simulator` (step 3 of [implementation plan](https://tier4.atlassian.net/browse/AJD-660))
- extract the vehicle model outside the `traffic_simulator` namespace to make it a step closer to moving it to the `simple_sensor_simulator` (step 4 of [implementation plan](https://tier4.atlassian.net/browse/AJD-660))

Changes introduced in this PR:
- `EgoEntitySimulation` class was introduced
  - It was extracted from the `traffic_simulator::entity::EgoEntity` class
  - Its functionality is mostly related to
    - managing the `SimulationModelInterface` object
    - managing the `concealer::Autoware` object
    - providing API to get the vehicle status (pose, action_status, etc)
- `EgoEntitySimulation` class files and `vehicle_model` directory were collected in a dedicated `vehicle_simulation` directory 
- Moved `EgoEntitySimulation` from the `traffic_simulator::entity::EgoEntity` class to `traffic_simulator::API`
  - It was done since `EgoEntitySimulation` will be moved to the `simple_sensor_simulator` in the next step
  - `traffic_simulator::API` is the place where ZeroMQ communication takes place and after moving `EgoEntitySimulation` to the  `simple_sensor_simulator`, it will be there where information about the Ego entity will be available
  - It forced adding ability to set `traffic_simulator::entity::EgoEntity` status externally using data from `EgoEntitySimulator`
 - Made `traffic_simulator::API` manage the `EgoEntitySimulation` object
   - This code will be moved to the `simple_sensor_simulator` in the next step 
   - This also forced the implementation of `requestSpeedChagne` methods in `traffic_simulator::API` rather than only forwarding them to the `EntityManager` since `EgoEntitySimulation` calls are needed in those methods
     - It will be removed after the next step is complete

## How to review this PR.
Several changes were introduced but a diff-based review should still be the most convenient way. Regression tests [have passed](https://tier4.atlassian.net/browse/AJD-660?focusedCommentId=126376).

## Others
[...]

PR #1184, opened 2024-02-01
Feature/consider tread in ego entity

hakuturu583:
# Description

## Abstract

- Enable consider tread when calculating matching length of EgoEntity.
- Fixed a problem with distance measurement in some scenarios ([e.g](https://evaluation.tier4.jp/evaluation/reports/486681a0-3903-5a3d-a59a-8423129b7a00/tests/c127e068-5273-5eae-9a4e-d98275660129/055ec2ad-ca6d-5f54-9ca0-64a749e0e7df/3ed30ec1-c158-52ca-91b4-fe73ba32e0b7?project_id=prd_jt)) that caused lane matching to fail.
- Removed externally_updated_status_, a member variable that stores the EnttiyStatus calculated by sensor_simulator, which existed only for EgoEntity.

It is not desirable to have many changes in a single pull request, but in this case, more than one change was made in a single pull request because externally_entity_status_ had to be removed in order to modify EgoEntity's lane coordinate system calculations.

### About lane coordinate system.

The lane coordinate system calculation is performed at least twice for EgoEntity.
The [first time is used to calculate the z-coordinates inside simple_sensor_simulator](https://github.com/tier4/scenario_simulator_v2/blob/b842a1a046a47a1142dc886c7b1b06335019b1cb/simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp#L400-L407), and [the second time is used to obtain the final lane coordinate system inside traffic_simulator.
](https://github.com/tier4/scenario_simulator_v2/blob/c37bb71e6bfe99247306de41fc7605588a7f58ba/simulation/traffic_simulator/src/api/api.cpp#L240-L246)

At the two locations mentioned above, the parameter for the length of the horizontal bar used to calculate the lane coordinates was fixed at 1.0 m.
Since lane coordinates are calculated from the intersection of this horizontal bar and the spline-completed curve of the lanelet centerline, the lane coordinate system cannot be calculated unless an intersection exists.

Screenshot from 2024-02-09 12-06-55

In this case, the scenario involved overtaking and avoiding a parked vehicle stopped on the left side of the road, so when the vehicle moved to the right inside the lane, it moved more than 1.0 m to the right and 2 to the right, so the lane coordinate system could not be calculated, and as a result, distance measurement in the longitudinal direction was impossible.

### About externally_updated_status

The externally_updated_status_ variable is used in [this pull request](https://github.com/tier4/scenario_simulator_v2/pull/998/files#r1250118429) to specify the It existed to temporarily store the EntityStatus calculated by simple_sensor_simulator, but other Entities were writing their updated results directly to the entity_status_ variable, creating confusion for developers.

Therefore, this variable has been removed in this pull request and changed to be set directly to entity_status_.

### About lane coordinate system.

The parameter for the length of the horizontal bar used for all lane coordinate calculations in EgoEntity was fixed at 1.0 m. 

The length of the horizontal bar for calculating the lane coordinate system after this pull request is imported is calculated from the following formula.

Let $L_m$ be the length of the abscissa used in the lane coordinate system calculation and the tread of the front wheels be  $t_f$ and the tread of the rear wheels be $t_r$.

$$L_m = max(t_r, t_f) * 0.5 + 1.0$$

### About externally_updated_status

Delete `EgoEntity::externally_updated_status_` member variable and use `EgoEntity::entity_status_` value.
If this change is not introduced, there will be a difference between the EgoEntityStatus inside traffic_simulator and simple_sensor_simulator, resulting in a discrepancy where the motion model on the simple_sensor_simulator simulator moves forward correctly but the `tf` topic issued from inside traffic_simulator does not.

# Destructive Changes

### About lane coordinate system.

Because the method of calculating the lane coordinate system is changed, the distance, etc. calculated in the lane coordinate system may change when this pull request is merged.

We tested the existing scenarios and confirmed that no regression exists for the existing test scenarios.

If you want to check regression test result, please check [here.](https://github.com/tier4/sim_evaluation_tools/issues/243)

# Known Limitations

### About lane coordinate system.

This pull request is not a major change to the process in order to emphasize backward compatibility with existing scenarios.

Therefore, the problem that exists in the current version of scenario_simulator_v2, "the lane coordinate system is calculated by extending a horizontal line vertically from the entity, so it is impossible to calculate the lane coordinate system for an entity placed at a large angle to the lane", has not been solved. This problem has not been solved.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -4,8 +4,9 @@
 
 Major Changes :race_car: :red_car: :blue_car:
 
-| Feature | Brief summary | Category | Pull request | Contributor |
-| ------- | ------------- | -------- | ------------ | ----------- |
+| Feature                 | Brief summary                                                        | Category            | Pull request                                                      | Contributor                                   |
+| ----------------------- | -------------------------------------------------------------------- | ------------------- | ----------------------------------------------------------------- | --------------------------------------------- |
+| Extend matching length. | Enable consider tread when calculating matching length of EgoEntity. | `traffic_simulator` | [#1181](https://github.com/tier4/scenario_simulator_v2/pull/1181) | [hakuturu583](https://github.com/hakuturu583) |
 
 Bug Fixes:bug:
 
--- a/simulation/simple_sensor_simulator/include/simple_sensor_simulator/sensor_simulation/sensor_simulation.hpp
+++ b/simulation/simple_sensor_simulator/include/simple_sensor_simulator/sensor_simulation/sensor_simulation.hpp
@@ -92,7 +92,7 @@ public:
   }
 
   auto attachPseudoTrafficLightsDetector(
-    const double current_simulation_time,
+    const double /*current_simulation_time*/,
     const simulation_api_schema::PseudoTrafficLightDetectorConfiguration & configuration,
     rclcpp::Node & node, std::shared_ptr<hdmap_utils::HdMapUtils> hdmap_utils) -> void
   {
--- a/simulation/simple_sensor_simulator/include/simple_sensor_simulator/vehicle_simulation/ego_entity_simulation.hpp
+++ b/simulation/simple_sensor_simulator/include/simple_sensor_simulator/vehicle_simulation/ego_entity_simulation.hpp
@@ -62,6 +62,7 @@ private:
 
 public:
   const std::shared_ptr<hdmap_utils::HdMapUtils> hdmap_utils_ptr_;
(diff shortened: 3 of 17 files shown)


hakuturu583:
Evaluator result : https://github.com/tier4/sim_evaluation_tools/issues/241

hakuturu583:
Please see also https://github.com/tier4/sim_evaluation_tools/issues/243

HansRobo (approved):

PR #1198, opened 2024-02-21
Feature/respawn entity

pawellech1:
# Description

## Abstract

This PR introduces the possibility of changing the position of ego manually using RViz GUI during the simulation in cpp mock scenarios.

Example scenario presenting the feature is included.

## Background

https://tier4.atlassian.net/browse/RJD-933)](https://tier4.atlassian.net/browse/RJD-933

## Details

- `API::respawn` method is implemented in this PR. 
  - inside the function UpdateEntityStatusRequest is used to change entity position in simple sensor simulator
  - after the relocation the route planned for this entity is canceled and replanned starting from the new position
  - finally concealer for the respawned entity is reengaged.
- For the purpose of `API::respawn` function:
  - `FieldOperatorApplicationFor<AutowareUniverse>::clearRoute()` method was implemented which triggers `/api/routing/clear_route` service
  - `EntityBase::getVehicleParameters` pure virtual method was implemented. For the entities other than VehicleEntity calling this method causes an exception.
  - new function `getParameterFromExistingNode` was implemented and used in `EgoEntitySimulation::makeSimulationModel` instead of `getParameter` function.
    - the reason for this change is the fact that during the respawn, multiple calls of `getParameter` function inside `EgoEntitySimulation::makeSimulationModel` were very time consuming (because of the need to construct and deconstruct multiple ROS2 nodes).
- `respawn_ego.cpp` scenario is implemented to present the future
  - in the scenario subscriber of `/initialpose` topic is implemented 
    - after receiveing the message from the topic, subscriber calls `API::respawn` passing newly received position and the initial ego goal as the arguments
  - using the `/initialpose` topic inside this scenario is just an example
    - some other topic or service could be used to trigger `API::respawn` call
    - using the `/initialpose` topic makes it possible to use RViz GUI “2d Pose Estimate” built in plugin

## Known Limitations

Currently only respawning of the `ego` entity is supported. Passing any other entity name to the function ends up with an exception.


mock/cpp_mock_scenarios/CMakeLists.txt
```diff
@@ -53,6 +53,8 @@ if(BUILD_CPP_MOCK_SCENARIOS)
   add_subdirectory(src/pedestrian)
   add_subdirectory(src/random_scenario)
   add_subdirectory(src/speed_planning)
+  # add_subdirectory(src/respawn_ego)
```
pawellech1, line 56: This line is commented because otherwise CI build step was not successful due to an error: https://github.com/tier4/scenario_simulator_v2/actions/runs/8005447800/job/21864899790#step:13:136.

I think this might be the general error related to cpp_mock_scenarios which use Autoware. Commenting this line and excluding the `respawn_ego` scenario from the build is the temporary solution making the CI pass.

mock/cpp_mock_scenarios/src/cpp_scenario_node.cpp
```diff
@@ -112,30 +113,14 @@ void CppScenarioNode::spawnEgoEntity(
...
-  });
-  while (!api_.asFieldOperatorApplication("ego").engaged() &&
-         !(initialized.load() && engaged.load())) {
-    // RCLCPP_INFO_STREAM(get_logger(), "Waiting for Autoware initialization...");
+    api_.updateFrame();
     std::this_thread::sleep_for(std::chrono::duration<double>(1.0 / 20.0));
   }
-  initialize_thread.join();
-  engage_thread.join();
 }
```
pawellech1, line 124: Here the idea of simplifying the "Waiting for Autoware initialization" state in CppScenarioNode is presented. Instead of creating two separate threads, one responsible for updating the simulation and the other responsible for engaging the concealer, I included both this actions in the current thread.

I do not see any reason why this part of code should not be simplified in this way, however if during the revision this simplification is found to be incorrect, please note it.

yamacir-kit:
@pawellech1 
Please resolve the conflict.

dmoszynski:
@yamacir-kit 
There was only one conflict and it was related to [pedestrian_entity.hpp](https://github.com/tier4/scenario_simulator_v2/pull/1198/files#diff-cbaae42f826417e4f510e4cb0384315eaedde09ba143925ceeed4a2416947c11). 
I've solved it quite simply and the changes made were not modified.
I believe that there is no need for re-run regressions.
Screenshot from 2024-03-25 09-50-52

yamacir-kit:
@pawellech1 
Sorry again and again. Please resolve the conflict.

Please be aware of the changes incorporated in 1210. In particular, the `EntityBase::getVehicleParameters` that this pull request is trying to introduce.

dmoszynski:
@yamacir-kit 
I've merged the master to the branch from this task and adapted to the changes. 
- However, I encountered issues - after merge `respawn` feature stopped working. 
- After a deep analysis, I found that it is necessary to publish the clock (ROS time) to Autoware.
- I added this publication, and on occasion rebuilt the code - https://github.com/tier4/scenario_simulator_v2/pull/1198/commits/1b9639fd8a0c5e2785a20bd16d3ae25e7aa125cc.

I'm also thinking whether it would not be better to develop the `respawn` in such a way that it would be performed inside the `updateFrame()` in the `API`. More precisely, for example, as a `Job`, so that ss2 modules update will be available during the respawn. The current solution is blocking and during the respawn the modules are not updated. I'm not sure if it is essential that the time required for `respawn` not cause the elapse of simulation time (iteration of frames). 

Let us know if this approach (of the original author) is as expected.If so, I will run regression tests again for the current version.

mock/cpp_mock_scenarios/CMakeLists.txt
yamacir-kit: Whether it is by OpenSCENARIO or a scenario written directly in C++, a build of Autoware is required to run a scenario containing Autoware on GitHub Actions.
In order to avoid too long execution times for GitHub Actions, and because the success of a scenario cannot be guaranteed since Autoware's behavior is not definitive, the repository is operated under the policy that tests involving Autoware are not performed on GitHub Actions.

yamacir-kit:
@dmoszynski 
The problem is that you are trying to perform a task in a single function that obviously cannot be completed in an instant.
`traffic_simulator` basically assumes that all operations can be performed without consuming simulation time. In other words, the functions other than `updateFrame` are the staging operations for the changes to occur in the next simulation frame, and `updateFrame` is the commit of the staged changes.

The `concealer` is usually responsible for the actual work, since the setup of Autoware is obviously not finished in an instant.
The `traffic_simulator` only instructs the `concealer` on the setup, while the actual work of the setup is performed by the `concealer` in an independent thread. (See the use of `task_queue.delay` in `field_opeartor_application_for_autoware_universe.cpp`)

>  I'm not sure if it is essential that the time required for respawn not cause the elapse of simulation time (iteration of frames).

The simulation start is simultaneous with the engage, but respawn is an event during the simulation, so it should be accompanied by a simulation time elapse.

dmoszynski:
@yamacir-kit 
> The problem is that you are trying to perform a task in a single function that obviously cannot be completed in an instant.
> traffic_simulator basically assumes that all operations can be performed without consuming simulation time. In other words, the functions other than updateFrame are the staging operations for the changes to occur in the next simulation frame, and updateFrame is the commit of the staged changes.

I understand this completely, I'm asking because the current solution is not of my invention and is a bit specific - does not cause elapsed time, that is, it blocks `updateFrame()` and I'm not sure if this is expected. I think we should avoid such blocking of `updateFrame()`, and some of the operations (`engageable()` check) should be done during `updateFrame()`. 

Although I must admit that now I've concluded that this check `engageble()` - the whole while loop, is unnecessary. We can assume that the `concealer` is responsible for the execution of `engage()` - that is, add the task to its queue and assume that it will be executed when possible. 

What do you think about this idea - [commit](https://github.com/tier4/scenario_simulator_v2/pull/1198/commits/bedf9eca420711feedebd9cd9650a76a4b11462c) (video below)? Unfortunately, this way we don't know if the "respawn" was successful/unsuccessful - we don't have a check on the `traffic_simulator` side to see if the `Ego` is `engaged()`. 

> The simulation start is simultaneous with the engage, but respawn is an event during the simulation, so it should be accompanied by a simulation time elapse.

From this I conclude that the check `engageble()` (the whole while loop) should be moved or removed. 

### Video
Video after change [commit](https://github.com/tier4/scenario_simulator_v2/pull/1198/commits/bedf9eca420711feedebd9cd9650a76a4b11462c) is added:

https://github.com/tier4/scenario_simulator_v2/assets/121798334/a5e8ecaf-8307-4fbe-8e7e-1cfaff2d9e22

yamacir-kit:
@dmoszynski 
The `engageable` is there to "engage Autoware and start the simulation as soon as Autoware is ready to engage". Therefore, when re-engaging during simulation, there is no problem to engage without checking engageable.

The `concealer`s `initialize`, `plan`, and `engage` have built-in confirmation that the execution of those functions will correctly transition the Autoware state. In other words, the `initialize` process confirms that the result will be a transition to `WaitingForRoute`, the `plan` process confirms that the result will be a transition to `WaitingForEngage`, and the `engage` process confirms that the result will be a transition to `Driving`. If the transition cannot be confirmed after a certain period of time, the `concealer` notifies an error and the simulation is shifted to shutdown as a failure. And `initialize`, `plan` and `engage` use `task_queue.delay`.

Therefore, there should be no problem to run `initialize`, `plan` and `engage` in succession without any confirmation. The transitions are automatically monitored by an independent thread, which is notified if a problem occurs, and the simulation proceeds as is if not. The success of respawn is checked consequently by the fact that no errors result.

I have not been able to check the video, but for the reasons above, the commit should be fine.
(It may be an issue with my environment, but I can't view the video on GitHub. It also fails to download.)

simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp
```diff
@@ -31,6 +31,16 @@ static auto getParameter(const std::string & name, T value = {})
...
 
+template <typename T>
+static auto getParameterFromExistingNode(
+  rclcpp::Node & node, const std::string & name, T value = {})
+{
+  node.declare_parameter<T>(name, value);
+  node.get_parameter<T>(name, value);
+
+  return value;
+}
```
yamacir-kit, line 42: Move it into makeSimulationModel (i.e., rewrite it as a function-local lambda expression). In that case, the variable name can be simply `get_parameter`.

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
```diff
@@ -221,6 +221,10 @@ class API
   bool despawn(const std::string & name);
   bool despawnEntities();
 
+  void respawn(
+    const std::string & name, const geometry_msgs::msg::PoseWithCovarianceStamped & new_pose,
+    const geometry_msgs::msg::PoseStamped & goal_pose);
```
yamacir-kit, line 226:
```suggestion
  auto respawn(
    const std::string & name, const geometry_msgs::msg::PoseWithCovarianceStamped & new_pose,
    const geometry_msgs::msg::PoseStamped & goal_pose) -> void;
```

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
```diff
@@ -319,13 +319,15 @@ class EntityManager
...
   FORWARD_TO_ENTITY(setBehaviorParameter, );
+  FORWARD_TO_ENTITY(setControlledBySimulator, );
   FORWARD_TO_ENTITY(setDecelerationLimit, );
   FORWARD_TO_ENTITY(setDecelerationRateLimit, );
   FORWARD_TO_ENTITY(setLinearJerk, );
   FORWARD_TO_ENTITY(setLinearVelocity, );
   FORWARD_TO_ENTITY(setMapPose, );
   FORWARD_TO_ENTITY(setTwist, );
   FORWARD_TO_ENTITY(setVelocityLimit, );
+  FORWARD_TO_ENTITY(getVehicleParameters, );
```
yamacir-kit, line 330: Please sort by lexicographical order.

simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp
```diff
@@ -120,6 +120,9 @@ class EntityBase
 
   virtual auto getBehaviorParameter() const -> traffic_simulator_msgs::msg::BehaviorParameter = 0;
 
+  virtual auto getVehicleParameters() const
+    -> const traffic_simulator_msgs::msg::VehicleParameters = 0;
```
yamacir-kit, line 124: The base class EntityBase should only have interfaces that are common to all entities. And EntityBase should not know what type of base class it will be. This function is odd in that EntityBase knows that it will be the base class for the Vehicle type and that the Vehicle type has a property called VehicleParameters.
Therefore, please remove this function and change the code that currently uses this function to use the capability query idiom for EntityBase types.

dmoszynski: You're totally right. I wasn't sure why `getVehicleParameters()` was added as part of this PR at all - I only suspected that it might be the result of some early discussion with @pawellech1. I've removed these changes completely - they are not needed for `respawn`.

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
dmoszynski: It has been removed.

dmoszynski:
@yamacir-kit 
Thank you so much for the review. 
- I've introduced all the requested changes. 
- I've run local tests, github CI is green. 

Taking a look at the commit with the changes I suppose that it is not necessary to re-run the regression tests - let me know what you think about it.

external/concealer/include/concealer/field_operator_application_for_autoware_universe.hpp
```diff
@@ -127,6 +129,7 @@ class FieldOperatorApplicationFor<AutowareUniverse>
     requestInitialPose("/api/localization/initialize", *this),
     // NOTE: /api/routing/set_route_points takes a long time to return. But the specified duration is not decided by any technical reasons.
     requestSetRoutePoints("/api/routing/set_route_points", *this, std::chrono::seconds(10)),
+    requestClearRoute("/api/routing/clear_route", *this),
```
yamacir-kit, line 132: Please sort by lexicographical order.

simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp
```diff
@@ -97,22 +97,28 @@ auto EgoEntitySimulation::makeSimulationModel(
   const traffic_simulator_msgs::msg::VehicleParameters & parameters)
   -> const std::shared_ptr<SimModelInterface>
 {
+  rclcpp::Node node{"get_parameter", "simulation"};
+  auto get_parameter = [](rclcpp::Node & node, const std::string & name, auto value = {}) {
+    node.declare_parameter<decltype(value)>(name, value);
+    node.get_parameter<decltype(value)>(name, value);
+    return value;
+  };
```
yamacir-kit, line 105:
```suggestion
  auto node = rclcpp::Node("get_parameter", "simulation");

  auto get_parameter = [&](const std::string & name, auto value = {}) {
    node.declare_parameter<decltype(value)>(name, value);
    node.get_parameter<decltype(value)>(name, value);
    return value;
  };
```

yamacir-kit (changes requested):
There were still a few things that bothered me and I'd like to fix them before merging.
There is no need to do another regression test on this change.

external/concealer/include/concealer/field_operator_application_for_autoware_universe.hpp
dmoszynski: Thank you for this, I had overlooked it...

simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp
dmoszynski: Indeed it looks better as there's no need to pass the “node” every each time. :bow:

yamacir-kit (approved):

dmoszynski (approved):
:)

PR #1194, opened 2024-02-19
Feature/default matching distance

hakuturu583:
# Description

## Abstract

- Unify the calculation criteria of threshold values used for lane coordinate system pose calculation to some extent for each Entity.
- Changed MiscObject/Pedestrian Entity to take crosswalks into account.

## Background

In #1189 and #1192, it was indicated that the existing lane coordinate system calculations were extremely complex, with different thresholds in various places, etc., and that it was presumed that proper processing was not being done.

## Details

- Unify the calculation criteria of threshold values used for lane coordinate system pose calculation to some extent for each Entity.
  - The matching distance values for the pose calculation in the lane coordinate system varied depending on various conditions. These were simplified and changed to basically use the matching distance determined for each Entity.
  - If the entity is the Ego or Vehicle entity, let $L_m$ be the length of the horizontal bar used in the lane coordinate system calculation and the tread of the front wheels be $t_f$ and the tread of the rear wheels be $t_r$. $$L_m = max(t_r, t_f) + 2.0$$
  - If the entity is the Pedestrian or MiscObject entity, let $L_m$ be the length of the horizontal bar used in the lane coordinate system calculation and width of the bounding box be $L_w$ $$L_m = L_w + 2.0$$

lane_pose_calculation

- Changed MiscObject/Pedestrian Entity to take crosswalks into account
  - They were set not to be taken into account for all Entities at spawn.

### Related issues / Pull requests
#1189 
This pull request is depends on #1192. This pull request become ready for review after merging #1192.

### Regression test result
https://github.com/tier4/sim_evaluation_tools/issues/244

# Destructive Changes

Although this Pull Request does not significantly change the processing flow of the existing algorithm, 100% backward compatibility does not exist because some of the thresholds have been changed.
However, it is assumed that the impact will be limited to a level that does not affect the existing scenario.
In fact, it has been confirmed that regression does not occur in existing scenarios.
See [test log](https://github.com/tier4/sim_evaluation_tools/issues/244) for details.

# Known Limitations

This Pull Request only changes the thresholds of the existing algorithm, and no major processing method changes exist.
Therefore, the specification that the lane coordinate system calculation fails if the horizontal bar extending from the Entity does not touch the curve at the center of the lane remains.

PR #1208, opened 2024-03-06
Fix/lanelet matching distance

TauTheLepton:
# Description

This fix adds the usage of default matching distance of each entity for `setEntityStatus` functions that are run by teleport action in open scenario.

## Abstract

Default lane matching distance depends on entity type and is used in various places in the code. The default distance was not used in all possible situations.

## Background

The default lane matching distance was added to the scenario simulator in PR #1194. However, the distance usage was not applied to all possible situations.

## Details

Default lane matching distance of each entity was not used in open scenario teleport action and the default distance of `1.0` was used. This fix adds the default lane matching distance to the `API::setEntityStatus` function, which is used in the open scenario teleport action.

## References

See #1194 for first default matching distance implementation.

# Destructive Changes

# Known Limitations

The same limitations to the lane matching as before apply. The difference is that the default lane matching distance in teleport action is used, so it is slightly bigger than before (which was `1.0`)

### DIFF
--- a/simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
+++ b/simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
@@ -331,6 +331,7 @@ public:
   FORWARD_TO_ENTITY_MANAGER(getRelativePose);
   FORWARD_TO_ENTITY_MANAGER(getStandStillDuration);
   FORWARD_TO_ENTITY_MANAGER(getTraveledDistance);
+  FORWARD_TO_ENTITY_MANAGER(getDefaultMatchingDistanceForLaneletPoseCalculation);
   FORWARD_TO_ENTITY_MANAGER(getV2ITrafficLight);
   FORWARD_TO_ENTITY_MANAGER(getV2ITrafficLights);
   FORWARD_TO_ENTITY_MANAGER(isEgoSpawned);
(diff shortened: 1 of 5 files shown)


simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
```diff
@@ -331,6 +331,7 @@ class API
   FORWARD_TO_ENTITY_MANAGER(getRelativePose);
   FORWARD_TO_ENTITY_MANAGER(getStandStillDuration);
   FORWARD_TO_ENTITY_MANAGER(getTraveledDistance);
+  FORWARD_TO_ENTITY_MANAGER(getDefaultMatchingDistanceForLaneletPoseCalculation);
   FORWARD_TO_ENTITY_MANAGER(getV2ITrafficLight);
   FORWARD_TO_ENTITY_MANAGER(getV2ITrafficLights);
   FORWARD_TO_ENTITY_MANAGER(isEgoSpawned);
```
hakuturu583, line 337: I think this function should not used by API users.
Please move it to private function.

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
hakuturu583: It is a very small change, so it is unnecessary to retry Autoware Evaluator.

hakuturu583 (approved):

PR #1209, opened 2024-03-07
Consider road slope in distance measurement and entity poses

HansRobo:
# Description

## Abstract

- consider z-axis in euclidean distance evaluation
- fit pitch angle of entity poses to HDMap 
- add `consider_pose_by_road_slope` flag to switch function
- update new scenarios and tests

Note: All the features can be toggled by `consider_pose_by_road_slope`.

## Background

One of the characteristics of `scenario_simualtor_v2` is that it is a simple and lightweight simulator.
Therefore, we have mainly dealt with simulations on a 2D plane without considering the Z direction or pitch/roll axes.

However, as the behavior on slopes is now being checked in autoware developing, simulation on a 2D plane alone is no longer able to cover the evaluation of Planning/Control, which is the main evaluation target of `scenario_simualtor_v2`.

So, we started to support simulation on 3D pose.

## Details

Note: I would like to thank @dmoszynski for his great efforts for this pull-request.

### Entity pose pitch-fitting to HDMap

As long as lane matching is successful, `scenario_simulator_v2` tries to constrain the pose on the HDMap.
Specifically, the situation is as follows.

- spawning entity to position: OpenSCENARIO `AddEntityAction`
- teleporting entity to position: OpenSCENARIO `TeleportAction`
- setting goal: OpenSCENARIO `AcquirePositionAction`
- setting route
  - OpenSCENARIO `FollowTrajectoryAction`
  - OpenSCENARIO `AssignRouteAction`

The z fitting to HDMap was already performed in `scenario_simulator_v2` until now. 
But the entity pose fitting was not complete until now: pitch and roll angle were not fitted to HDMap. 
In this pull-request, pitch-fitting is implemented.  

image

With OpenSCENARIO `WorldPosition` / `RelativeWorldPosition`  / `RelativeObjectPosition`, `scenario_simulator_v2` will be performing both of z-fitting and pitch-fitting.
With OpenSCENARIO `LanePosition`, `scenario_simulator_v2` will be performing only pitch fitting.
( Other position types are not supported in `scenario_simulator_v2` now )

The pitch-fitting is also affected to bounding-box-based distance evaluation(freespace=true), because the orientation of entity bounding-box can be changed by pitch-fitting.

### Distance evaluation

The euclidean distance evaluation in OpenSCENARIO `DistanceCondition` / `RelativeDistanceCondition` / `ReachPositionCondition` is performed in 2D plane previously.
In this pull-request, I implemented 3D euclidean dintance evaluation.

image

## References

[INTERNAL INVESTIGATION LINK](https://tier4.atlassian.net/browse/RJD-770)

[Regression Test: Pass](https://github.com/tier4/sim_evaluation_tools/issues/251)

# Destructive Changes

No API breaking changes, but some of API functions have additional options with backward compatibility.
Some of scenarios or APIs behaviors are changed as mentioned in detail section.
There should be almost no difference on a flat map, but there are some changes, so you may need to modify the scenario.
However, these changes correct what was wrong in the past, so it is recommended that you accept them.
If you really want to use the old behavior, please use the version before this pull-request or specify consider_pose_by_road_slope=False.

# Known Limitations

## No roll-fitting to HDMap

In this pull-request, orientation fitting is implemented for pitch angle only.
This is because it follows the implementation of `simple_planning_simulator` in `autoware.universe`.

## No orientation-fitting to HDMap in `OccupancyGridSensor`

The GridMap output from `scenario_simulator_v2` is always horizontal regardless of the road slope.
This is because current implementation of autoware.universe does not consider the roll and pitch axes.


HansRobo:
This is re-opened pull-request of https://github.com/tier4/scenario_simulator_v2/pull/1103

common/math/geometry/src/spline/hermite_curve.cpp
```diff
@@ -279,7 +279,9 @@ const geometry_msgs::msg::Vector3 HermiteCurve::getTangentVector(double s, bool
   return vec;
 }
 
-const geometry_msgs::msg::Pose HermiteCurve::getPose(double s, bool denormalize_s) const
+// @note fill_pitch is introduced for backward-compatibility.
```
hakuturu583, line 282:
```suggestion
/// @note fill_pitch is introduced for backward-compatibility.
```

openscenario/openscenario_interpreter/src/syntax/distance_condition.cpp
```diff
@@ -101,6 +107,16 @@ auto DistanceCondition::distance(const EntityRef & triggering_entity) const -> d
...
 }
 
+// @todo: after checking all the scenario work well with consider_z = true, remove this function and use std::hypot(x,y,z)
+static double hypot(const double x, const double y, const double z, const bool consider_z)
+{
+  if (consider_z) {
+    return std::hypot(x, y, z);
+  } else {
+    return std::hypot(x, y);
+  }
```
hakuturu583, line 117:
```suggestion
  return consider_z ? std::hypot(x, y, z) : std::hypot(x, y);
```

openscenario/openscenario_interpreter/src/syntax/reach_position_condition.cpp
```diff
@@ -47,29 +53,39 @@ auto ReachPositionCondition::description() const -> String
...
 }
 
+// @todo: after checking all the scenario work well with consider_z = true, remove this function and use std::hypot(x,y,z)
+static double hypot(const double x, const double y, const double z, const bool consider_z)
+{
+  if (consider_z) {
+    return std::hypot(x, y, z);
+  } else {
+    return std::hypot(x, y);
+  }
```
hakuturu583, line 63:
```suggestion
  return consider_z ? std::hypot(x, y, z) : std::hypot(x, y);
```

openscenario/openscenario_interpreter/src/syntax/relative_distance_condition.cpp
```diff
@@ -112,6 +123,16 @@ auto RelativeDistanceCondition::distance<
...
 }
 
+// @todo: after checking all the scenario work well with consider_z = true, remove this function and use std::hypot(x,y,z)
+static double hypot(const double x, const double y, const double z, const bool consider_z)
+{
+  if (consider_z) {
+    return std::hypot(x, y, z);
+  } else {
+    return std::hypot(x, y);
+  }
```
hakuturu583, line 133:
```suggestion
  return consider_z ? std::hypot(x, y, z) : std::hypot(x, y);
```

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
```diff
@@ -542,8 +540,17 @@ class EntityManager
             lanelet_pose) {
           entity_status.lanelet_pose = *lanelet_pose;
           entity_status.lanelet_pose_valid = true;
+          // @note fix z, roll and pitch to fitting to the lanelet
```
hakuturu583, line 543:
```suggestion
          /// @note fix z, roll and pitch to fitting to the lanelet
```

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
```diff
@@ -542,8 +540,17 @@ class EntityManager
...
+          static bool consider_pose_by_road_slope =
+            getParameter<bool>("consider_pose_by_road_slope", false);
```
hakuturu583, line 545:
```suggestion
          bool consider_pose_by_road_slope = getParameter<bool>("consider_pose_by_road_slope", false);
```

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
```diff
@@ -542,8 +540,17 @@ class EntityManager
...
+          if (consider_pose_by_road_slope) {
+            entity_status.pose = hdmap_utils_ptr_->toMapPose(*lanelet_pose).pose;
+          } else {
+            entity_status.pose = pose;
+          }
```
hakuturu583, line 550:
```suggestion
          if (getParameter<bool>("consider_pose_by_road_slope", false)) {
            entity_status.pose = hdmap_utils_ptr_->toMapPose(*lanelet_pose).pose;
          } else {
            entity_status.pose = pose;
          }
```

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp
hakuturu583: This line is executed when the Entity spawns, so it is not executed every frame.

HansRobo: Oh, I see.
Then, it's reasonable.

hakuturu583 (approved):

PR #1216, opened 2024-03-22
Implement `DistanceCondition` / `RelativeDistanceCondition` for `shortest` of `RoutingAlgorithm`

HansRobo:
# Description

## Abstract

In this pull-request, I implemented a lane change enabled routing algorithm, as `shortest` of OpenSCENARIO `RoutingAlgorithm` for `DistanceCondition` / `RelativeDistanceCondition`.

## Background

To resolve the difference of two lane positions, the `shortestPath` of `lanelt2` with a no lane change option is used as a routing algorithm in current implementation of `DistanceCondition` and `RelativeDistanceCondition`.

Although this implementation has been used in many scenarios, it returns distances that are not intuitive or cannot be evaluated in some use cases.
One of the typical use cases is shown in the figure below.
When measuring the distance from a red car to a blue car, intuitively we assume that the distance will be measured along the red arrow, but in reality the distance will be measured along the blue arrow.

image

### Behavior in the specific situation

image

#### RoutingAlgorithm: `undefined`
`undefined` does not change lanes, so the distances will have to go around B and C to get there.
Since `undefined` does not change lanes, the route for distance measurement goes around 100,101,102 and 11,12 to reach the blue car.
In the case of this figure, distances would be evaluated as follows
- longitudinal distance: 73
- lateral distance: -2

Note: The value of lateral distance has a negative value, which is known to be a violation of the OpenSCENARIO standard. (`shortest`'s implementation fixes it.)

#### RoutingAlgorithm: `shortest`

By changing lanes, `shortest` takes the most intuitive route in this figure, connecting 2 and 12 directly.

In the case of this figure, distances would be evaluated as follows

- longitudinal distance: 4
- lateral distance: 6

### Tests

In `ByEntityCondition.EntityCondition.DistanceCondition.Shortest.yaml`, I set up 24(3 * 2 * 2 * 2) tests.

image

- 3 situations
  - NPC1: routable without any lane changes by looping, but there is shorter route with a lane change
  - NPC2: in adjacent lane (for edge case)
  - NPC3: in in adjacent lane and behind ego (for edge case)
-  2 Conditions
   - `DistanceCondition`
   - `RelativeDistanceCondition`
- 2 `freespace` options for distance evaluation
  - true
  - false
- 2 axes in lane coodinate system
  - lateral
  - longitudinal  

## References

[OpenSCENARIO XML 1.3 RoutingAlgorithm](https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_XML/latest/generated/content/RoutingAlgorithm.html)

[regression test result](https://github.com/tier4/sim_evaluation_tools/issues/262): OK 

# Destructive Changes

If nothing is specified for `RoutingAlgorithm` in the scenario, `undefined` will be used, as defined in OpenSCENAIRO standard.

> Algorithm for path selection/calculation between two positions across roads. Only relevant, if CoordinateSystem is "road"/"lane". Default value if omitted: "undefined".
https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_XML/latest/generated/content/DistanceCondition.html

The legacy implementation is used as this "undefined" behavior, so there are no breaking changes to existing scenarios.

# Known Limitations

N/A


simulation/traffic_simulator/src/hdmap_utils/hdmap_utils.cpp
```diff
@@ -1455,18 +1491,63 @@ auto HdMapUtils::getLongitudinalDistance(
...
+      auto next_lanelet_itr = std::find_if(
+        next_lanelet_ids.begin(), next_lanelet_ids.end(),
+        [next_lanelet](const lanelet::Id & id) { return id == next_lanelet; });
+      return next_lanelet_itr == next_lanelet_ids.end();
+    } else {
+      return false;
+    }
+  };
+  for (unsigned int i = 0; i < route.size(); i++) {
+    if (i < route.size() - 1 && with_lane_change(allow_lane_change, route[i], route[i + 1])) {
```
hakuturu583, line 1514: I suggest to use

```
  for (auto itr = std::adjacent_find(route.begin(), route.end(),
         [](const auto &, const auto &) { return true; });
```

hakuturu583 (changes requested):

hakuturu583 (changes requested):
I believe that this is a pull request to add important functionality related to distance measurement, so please also add documentation for each measurement method.

simulation/traffic_simulator/src/hdmap_utils/hdmap_utils.cpp
HansRobo: I don't think `std::adjacent_find` is match for this usecase and introduce bug especially with multiple lane change route.

hakuturu583: Sorry, I misunderstood the usage of std::adjacent_find

```
if (auto itr = std::adjacent_find(route.begin(), route.end(),
       [](const auto & waypoint_a, const auto & waypoint_b) { `Functions return true if the ` }; 
   itr != route.end()){
   }
```
```

hakuturu583 (approved):

PR #1217, opened 2024-03-27
Feature/follow trajectory in do nothing plugin

hakuturu583:
# Description

## Abstract

Start supporting `API::requestFollowTrajectory` function in do nothing plugin.

## Background

The behavior of the API::requestFollowTrajectory function when using the behavior tree plugin is useful in that it can take into account various constraints, but it is difficult to use in rosbag_to_scenario where complete observation information is not available due to the high failure rate. 

## Details

We implemented FollowTrajectoryAction, whose function is limited to do_nothing_plugin, which ignores the laws of physics.

## References

[OpenSCENARIO document.](https://releases.asam.net/OpenSCENARIO/1.0.0/Model-Documentation/content/FollowTrajectoryAction.html)

# Destructive Changes

> [Optional] If no destructive change exists, you can skip this section.  
> Include a description of the changes and a migration guide and send the pull request with a bump major label. (Example : https://github.com/tier4/scenario_simulator_v2/pull/1139)  
> Otherwise, skip the "Destructive Changes" section and make sure this pull request is labeled `bump minor` or `bump patch`.  

# Known Limitations

> [Optional] If there are no known limitations that you can think of, you can skip this section. If there are any limitations on the features or fixes added by this pull request, delete this sentence and clearly describe them.
> For example, the lane matching algorithm currently (1/25/2024) employed is unable to match Entity that is heavily tilted with respect to the lane, and it is difficult to throw an exception.  
> If the developer is aware of the existence of such problems or limitations, describe them in detail. The problems or limitation should be listed as an Issue on GitHub and a link to it should be provided in this section.


simulation/do_nothing_plugin/include/do_nothing_plugin/plugin.hpp
```diff
@@ -74,13 +71,19 @@ public:                                                                \
...
+  DEFINE_GETTER_SETTER(HdMapUtils,         std::shared_ptr<hdmap_utils::HdMapUtils>,                         hdmap_utils_)
+  DEFINE_GETTER_SETTER(PolylineTrajectory, std::shared_ptr<traffic_simulator_msgs::msg::PolylineTrajectory>, polyline_trajectory)
+  DEFINE_GETTER_SETTER(Request,            traffic_simulator::behavior::Request,                             request)
+  DEFINE_GETTER_SETTER(StepTime,           double,                                                           step_time_)
+  DEFINE_GETTER_SETTER(UpdatedStatus,      std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus>,    updated_status_)
   // clang-format on
 #undef DEFINE_GETTER_SETTER
+
+  void checkPolylineTrajectory();
+  void followPolylineTrajectory();
```
yamacir-kit, line 86: These functions are for implementing DoNothingBehavior::update, so they do not need to be member functions. It is sufficient that they are local functions in the .cpp file.

yamacir-kit, line 86: These functions are for implementing `DoNothingBehavior::update`, so they do not need to be member functions. It is sufficient that they are local functions in the `.cpp` file.

mock/cpp_mock_scenarios/src/follow_trajectory/follow_polyline_trajectory_with_do_nothing_plugin.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <rclcpp/rclcpp.hpp>
+#include <traffic_simulator/api/api.hpp>
+#include <traffic_simulator_msgs/msg/behavior_parameter.hpp>
+#include <traffic_simulator_msgs/msg/polyline.hpp>
+#include <traffic_simulator_msgs/msg/vertex.hpp>
+
+// headers in STL
+#include <memory>
+#include <string>
+#include <vector>
```
yamacir-kit, line 29: The term `STL` is not a term that exists in the C++ standard. Some of the standard libraries are still conventionally called STL, but at least some of the libraries you call "STL" definitely include `memory` that is not included in STL, even by convention.If you

If you think that "STL" is a term referring to the C++ standard library, you are wrong. Please do not use this terminology.

simulation/do_nothing_plugin/src/plugin.cpp
```diff
@@ -13,15 +13,131 @@
...
+            interpolate_entity_status(
+              (current_time_ + step_time_ - trajectory->base_time - timestamp_i) /
+                (timestamp_i_1 - timestamp_i),
+              trajectory->shape.vertices[i], trajectory->shape.vertices[i + 1]),
+            getHdMapUtils())));
+        break;
+      }
+    }
+  }
+  if (!updated_status_) {
```
yamacir-kit, line 138: This conditional expression seems to always evaluate to false. It is very likely a bug.

simulation/do_nothing_plugin/src/plugin.cpp
```diff
@@ -13,15 +13,131 @@
...
+  }
+}
+
+void DoNothingBehavior::followPolylineTrajectory()
+{
+  updated_status_ = std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus>();
+  checkPolylineTrajectory();
+
+  if (const auto trajectory = getPolylineTrajectory()) {
+    for (size_t i = 0; i < trajectory->shape.vertices.size() - 1; i++) {
```
yamacir-kit, line 71: The algorithm "iterate over vertices by index and break when it finds one that matches the condition" seems very bug-prone. This block should be written using std::adjacent_find.

yamacir-kit (changes requested):

mock/cpp_mock_scenarios/src/follow_trajectory/follow_polyline_trajectory_with_do_nothing_plugin.cpp
hakuturu583: Fixed in 17e1aa657fea28861cb18583b7579189bce10c29

simulation/do_nothing_plugin/include/do_nothing_plugin/plugin.hpp
hakuturu583: Fixed in https://github.com/tier4/scenario_simulator_v2/pull/1217/commits/395ee4658d19ad9dce2ae7cf9a0f4a6ca2cd9cca

simulation/do_nothing_plugin/src/plugin.cpp
hakuturu583: Fixed in https://github.com/tier4/scenario_simulator_v2/pull/1217/commits/fee36c909165f967703b5eace2f36f3e7070b47c

PR #1223, opened 2024-04-04
Delete workflow feature and add alternative bash script

HansRobo:
# Description

## Abstract

- Delete the workflow feature from `scenario_test_runner` to simplify it
- Add a bash script to replace workflow feature

## Background

The workflow feature was considered necessary at the initial stage of `scenario_simulator_v2`.
However, over time, we found that the workflow feature did not provide useful functionality to justify the added complexity of the data flow.
And I'm removing this feature in this pull-request as a simple script can replace it.

### usage for new script

```
./workflow.sh <workflow file> <other options for scenario_test_runner>
```

### workflow file example

A file with a scenario path written on each line is valid.
It is passed to the scenario specification part of `scenario_test_runner`, so special commands such as `find-pkg-share` can be used.

```txt
$(find-pkg-share scenario_test_runner)/scenario/all-in-one.yaml
$(find-pkg-share scenario_test_runner)/scenario/minimal.yaml
```

#### script exit status

The script's exit status is 1 if even one scenario fails, and 0 if all are successful.

## References

[Regression Test Result](https://github.com/tier4/sim_evaluation_tools/issues/264): OK

# Destructive Changes

Workflow functionality has been removed.
Users must create their own replacement functionality.
A simple sample is included in this pull request for your reference.

# Known Limitations

N/A


.github/workflows/workflow.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+FILE_PATH="$1"
```
yamacir-kit, line 3: If it is not an environment variable, use lower_snake_case for the variable name.

.github/workflows/workflow.sh
```diff
@@ -0,0 +1,26 @@
...
+
+exit_status=0
+
+while IFS= read -r line
+do
+    ros2 launch scenario_test_runner scenario_test_runner.launch.py scenario:="$line" "$@"
+    ros2 run scenario_test_runner result_checker.py /tmp/scenario_test_runner/result.junit.xml
+    cmd_exit_status=$?
+    if [ $cmd_exit_status -ne 0 ]; then
+        echo "Error: caught non-zero exit code（code: $cmd_exit_status)"
```
yamacir-kit, line 21:
```suggestion
        echo "Error: caught non-zero exit status（code: $cmd_exit_status)"
```

.github/workflows/workflow.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+FILE_PATH="$1"
+
+# Arguments other than file path are passed to the command's arguments
+shift 1
+
+if [ ! -f "$FILE_PATH" ]; then
+    echo "No such file: $FILE_PATH"
+    exit 1
```
yamacir-kit, line 10: If there is no special reason, change the indentation to 2 spaces wide to match the other languages in the repository.

.github/workflows/workflow.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+exit_status=0
+
+while IFS= read -r line
+do
+    ros2 launch scenario_test_runner scenario_test_runner.launch.py scenario:="$line" "$@"
+    ros2 run scenario_test_runner result_checker.py /tmp/scenario_test_runner/result.junit.xml
+    cmd_exit_status=$?
```
yamacir-kit, line 19:
```suggestion
    command_exit_status=$?
```

.github/workflows/workflow.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+FILE_PATH="$1"
+
+# Arguments other than file path are passed to the command's arguments
+shift 1
+
+if [ ! -f "$FILE_PATH" ]; then
```
yamacir-kit, line 8:
```suggestion
if [ ! -f "$FILE_PATH" ]
then
```

yamacir-kit (changes requested):

.github/workflows/workflow.sh
HansRobo: fixed in https://github.com/tier4/scenario_simulator_v2/pull/1223/commits/bbcc4a618d1651a303931d0592e9026937614a0f

.github/workflows/workflow.sh
HansRobo: fixed in https://github.com/tier4/scenario_simulator_v2/pull/1223/commits/de2d26ed92f7df11a520b35a01949fcc7de31426

.github/workflows/workflow.sh
HansRobo: fixed in https://github.com/tier4/scenario_simulator_v2/pull/1223/commits/a35a58bd8e539793291e7eac702c4caae10d129e

.github/workflows/workflow.sh
HansRobo: fixed in https://github.com/tier4/scenario_simulator_v2/pull/1223/commits/5186ac984da5e69ccbe41d59f9ffb7fe80ede115

yamacir-kit (approved):

PR #1226, opened 2024-04-10
fix(follow_trajectory_action): adapt to work with considering slopes

dmoszynski:
# Description
#### Fixes
- fix NaN time in the EntityStatus at FollowTrajectoryAction start,
- fix matching distance for `toLaneletPose()` in FollowTrajectoryAction,
- fix matching distance in EgoEntitySimulation and EntityBase,
- fix FollowTrajectoryAction for slopes (use the distance along the lanes and take into account the pitch),
  - note: I’ve made sure that FollowTrajectoryAction works properly also outside lanes, as it has been until now.
## Background

I’ve adapted FollowTrajectoryAction to get the Passed result for scenario [RoutingAction.FollowTrajectoryAction-autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml).

## NaN time - details
- The issue was related to the NaN time in the EntityStatus at FollowTrajectoryAction start.
- This occurred due to the missing setting of the time equal to 0 in Entities statuses at the start of the simulation (calling startNpcLogic()) - [commit](https://github.com/tier4/scenario_simulator_v2/commit/62b9867a9084a3e191f74c6889d8159f79532ea4) with the fix, most significant changes:
   - [Add set current_time in startNpcLogic() for EntityManager](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-2a631ea36836a3ae21fbc2176a966e6969c518653585de530b78e6b12fd2c03e)
   - [Add set current_time in startNpcLogic() for EntityBase](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-3470b455c589da1b490a6087dee3a95bffc20035b732177650b76b4ddd2bc004)

## Matching distance - details
- The issue was related to the failure to match the lane when changing lanes.
- The issue has arisen as a result of the changes involved in the unification of the matching_distance values when matching Entity position to lane - [source PR](https://github.com/tier4/scenario_simulator_v2/pull/1194). 
- This occurred due to missing adaptation of FollowTrajectoryAction to the new approach, to be more precise matching_distance on the BehaviorTree side was different than in TrafficSimulator - [commit](https://github.com/tier4/scenario_simulator_v2/commit/4d92606e3e85cb51d4c8bf6564c96f1b7e237125) with the fix, most significant changes:

  - [Fix matching distance in FollowTrajectoryAction for pedestrian](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-e14cb8f1a98cad765e5c9802d90ba3b255ea4782e9640d1b560c03e76e910cb4)
  - [Fix matching distance in FollowTrajectoryAction for vehicle](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-0f9b92a171a25454e7139d2d85c0bf63713f5bb187120eb22d79e8daa03fe985)
  - [Fix matching distance in FollowTrajectoryAction for ego](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-9725e59d0fe04763f7ac3e2df38fc9fc24bd0319af80b4dffee45f0bc3660ed4)
  - [Fix matching distance in fillLaneletPose() for EntityBase](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-3470b455c589da1b490a6087dee3a95bffc20035b732177650b76b4ddd2bc004)
  - [Fix matching distance in getMatchedLaneletPoseFromEntityStatus() for EgoEntitySimulation](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-a70b23781f2bdd4b04e18eb52c51dd37fd686738ec2b720c1ec3c5fc6844e462)

## Slope adaptation - details
After making the above changes, FollowTrajectoryAction successfully arrives at all waypoints, but there is a trouble with having enough accuracy when stopping at the last one.
- This issue has arisen as a result of changes involved in the introduction of consideration of the slope in the position of Entity - [source PR](https://github.com/tier4/scenario_simulator_v2/pull/1209).
- I investigated this, and after setting `consider_pose_by_road_slope` to false, FollowTrajectoryAction achieves sufficient accuracy, which confirmed that [these slope-related](https://github.com/tier4/scenario_simulator_v2/pull/1209) changes are causing this issue.
   - Note: An additional code modification was required, related to temporarily revert [this change](https://github.com/tier4/scenario_simulator_v2/pull/1209/files#diff-da44510bdbbba766d1ba47318640cfd8bcff2e350eafe3d77d364bfbf70e25cd) - unfortunately, it seems that `consider_pose_by_road_slope` has no effect on [HdMapUtils::toMapPose()](https://github.com/tier4/scenario_simulator_v2/blob/918f2af746a88c46fedd4ab2630f85f2c7450fd5/simulation/traffic_simulator/src/hdmap_utils/hdmap_utils.cpp#L1413). Despite setting `consider_pose_by_road_slope` to false, orientation is still calculated with the slope taken into account.

To better understand the issue, I’ve developed an additional scenario "Scenario with slopes" in which there are higher and longer slopes. FollowTrajectoryAction is triggered after 1 second from the start of the scenario and its waypoints form a Polyline is similar to the one marked in red in the figure.
image
[Link to scenario with slopes and map files](https://drive.google.com/drive/folders/1Ey_6rQbDLLQwSaNG1aXrV7JfYXoX_6fu).

#### 1.1 Before changes - Scenario with slopes - NPC:
As can be seen, the NPC does not move on the lanelet, but moves along the Polyline created by waypoints directly, so that sometimes it is above the lanelet and sometimes below it. The pitch is all the time equal to 0.
In such a case, the calculation of the remaining distance using the euclidean distance works correctly - FollowTrajectoryAction chooses the acceleration well, as a result of which the NPC, after accelerating, moves at a constant speed from which it brakes at the last waypoint.

 https://github.com/tier4/scenario_simulator_v2/assets/121798334/098b2a2e-c26d-43d4-93c4-bce37f13381e
   
#### 1.2 Before changes - Scenario with slopes - EGO
As can be seen, the position (in Oz axis) of the Ego, compared to the case for the NPC, is aligned to the lanelet. Because of this, the position on the output of the action after the FollowTrajectoryAction step is different from that on the input of the action in the next step. As a result of that, Ego has a position other than the expected (set) one, FollowTrajectoryAction continues to set accelerations all the time, and Ego does not have a constant speed. The pitch (on FollowTrajectoryAction side)  is all the time equal to 0.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/9bc0f7d6-abaf-403d-9314-67147ebadd33

#### 1.3 Before changes -  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - NPC
As can be seen, in this case FollowTrajectoryAction works well, and pitch slopes and Oz axis changes are so small that they are unnoticeable.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/722a569b-d699-4b41-87d8-9e723489354e

##### 1.4 Before changes -  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - EGO
However, for Ego, issues arise. They are caused by the same reason as described in 1.2 - the position at the output of FollowTrajectoryAction is different from that at the input in the next step. Ego's velocity changes unexpectedly when changing lanes (where there are differences in the Oz axis and pitch). Because of these issues, FollowTrajectoryAction does not work properly and has a large inaccuracy when stopping.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/c546b845-0ffb-47fc-a8dd-3867c1b8e328

##### Most significant changes:
- use `distance_along_lanelet` for the calculations of the remaining distance
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L73-L90
- use `fill_lanelet_data_and_adjust_orientation` for lanelet position filling, taking pitch from lanelet and applying it to entity orientation
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L91-L113
- calc `desired_velocity` vector using pitch along lanelet (if possible) and yaw on the target
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L378-L399
- consider pitch in `current_velocity` vector
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L419-L428
- use `desired_velocity` to calc entity orientation
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L583-L596
- after the step, match the orientation to the lanelet
https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L598-L603

##### Results:
- For NPC FollowTrajectoryAction works correctly and as expected.
- I noticed that Ego executes the FollowTrajectoryAction in the scenario in a different way than the NPC - despite the fact that it is then controlled by the same method.
- The changes related to adaptation to consider slopes were not sufficient to get Passed.

#### 2.1 Adaptation for slopes - Scenario with slopes - NPC: 
As can be seen, FollowTrajectoryAction moves the NPC along the lanalet taking into account its slope - as expected. Accelerations are also calculated correctly, ensuring that the NPC reaches a constant speed and stops with sufficient accuracy.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/f660c54f-6c09-4897-9d59-3639447d5d62

#### 2.2 Adaptation for slopes - Scenario with slopes - EGO
As can be seen, despite the improvement for Npc, the issue for Ego still exists - speeds change in unexpected ways - after the FollowTrajectoryAction step, the input position is different than the output position in the previous step.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/dc867fc5-19ae-4595-b069-998b3346f8e0

#### 2.3 Adaptation for slopes  -  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - NPC
As can be seen now, when the NPC is aligned in the Oz axis to the lanelet and the pitch is taken into account, there are speed changes during lane changes here. This is because the waypoints are in other lanes, so that the distance along the lanelet cannot be calculated, and the euclidean distance is used. (Here the position in Oz is different on both lanes - I checked it, however, I did not take a screen shot...). However, here, for the NPC, after the FollowTrajectoryAction step, the input position is the same as the output position in the previous step.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/e182e592-ed60-46f1-8fe1-351e5eac39a8

#### 2.4 Adaptation for slopes-  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - EGO
As can be seen, the Ego behaves similarly to the NPC - here there are also speed changes caused by the position in Oz. However, for Ego, after the FollowTrajectoryAction step, the input position is different than the output position in the previous step. For this reason, FollowTrajectoryAction cannot achieve the expected accuracy.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/5d54c559-8a50-461b-939b-6c99ecc165ee

##### Most significant changes:
- [getZ() - in SimModelDelaySteerAccGeared ](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-1e688465cc7fd948006ce3e524f980a5f93d90968069d6d300f77652790c1d9e)
- [overwrite Oz position in model and read it - in EgoEntitySimulator](https://github.com/tier4/scenario_simulator_v2/pull/1226/files#diff-a70b23781f2bdd4b04e18eb52c51dd37fd686738ec2b720c1ec3c5fc6844e462)

##### Results:
- The issue caused that the read pose was different from the written pose (which should not happen, as the pose is overwritten instead of updated)
- Now FollowTrajectoryAction works properly for slopes as well, both for NPCVehicle and for Ego.

#### 3.1 Fix for Ego - Scenario with slopes - NPC: 

The same as 2.1.

#### 3.2 Fix for Ego - Scenario with slopes - EGO
As can be seen, now the Ego behaves exactly the same as the NPC - after the FollowTrajectoryAction step, the input position is the same as the output position in the previous step. FollowTrajectoryAction ensures stopping at the last point with sufficient accuracy.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/cf70f108-a132-4ba4-91be-5cd10c57c664

#### 3.3 Fix for Ego  -  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - NPC

The same as 2.3.

#### 3.4 Fix for Ego -  [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - EGO
As can be seen, now the Ego behaves exactly the same as the NPC - after the FollowTrajectoryAction step, the input position is the same as the output position in the previous step. FollowTrajectoryAction ensures stopping at the last point with sufficient accuracy.

https://github.com/tier4/scenario_simulator_v2/assets/121798334/9cb4a5fd-2c79-4864-9013-8516863838d3

# Destructive Changes

### [Oz position in SimModelInterface](https://github.com/tier4/scenario_simulator_v2/pull/1226/commits/46490d6dd7239416c39905652878692b19dee53b)
- This change is necessary for FollowTrajectoryAction to work properly. Entities move along lanelets, where pitch and Oz are taken into account. If FollowTrajectoryAction is to overwrite Ego positions, the position in Oz cannot just be ignored.
- Currently, the virtual `getZ()` method has been implemented, which returns 0 for every model except the one currently in use.
- The position in Oz is set only when the position is overwritten, it is not calculated in any way. To be more precise, if the `update()` in SimModelInterface is used, everything still works on 2D, and the position in Oz is only used during `overwrite()` - that is, only by FollowTrajectoryAction.
- I believe that all SimModelInterface should be adapted to 3D, due to the fact that all the rest of the code in ss2 currently uses 3D and Autoware uses 3D.

### Using [distance along](https://github.com/tier4/scenario_simulator_v2/blob/dfb711faa5ca7f1eafc0f3157076bfca519445f7/simulation/traffic_simulator/src/behavior/follow_trajectory.cpp#L224) `lanelet2` if possible instead of euclidean distance
- This seems to be necessary for FollowTrajectoryAction to work properly.
- However, this introduces potential issues when the position on the lanelet is not unique, i.e., for example, when passing through lanelet connections, the entity is once on the previous one once on the one in front, and in the next step it is again on the previous one. For example, the following lanelet_ids: {1,2}, each of length 50, assume that in step (1) the Entity is at lanelet_id=1 and s=49.9, in the next step (2) it will be at lanelet_id=2 and s=0. 0, but in step (3) it may be again at lanelet_id=1 and s=49.9. This happened during my tests and it is worth to ensure that such situations do not occur, as it causes the distance along the lanelet to change in unexpected ways.
- It is also important to take into account the issue of calculating the distance during lane changes, which cause the distance to change in unexpected ways.


simulation/simple_sensor_simulator/include/simple_sensor_simulator/vehicle_simulation/vehicle_model/sim_model_delay_steer_acc_geared.hpp
```diff
@@ -62,6 +62,7 @@ class SimModelDelaySteerAccGeared : public SimModelInterface
     VX,
     STEER,
     ACCX,
+    Z,
```
yamacir-kit, line 65: The vehicle_model in scenario_simulator_v2 is a hard copy of the one included in autowarefoundation/autoware. This is done to keep the differences between scenario_simulator_v2 and simple_planning_simulator small. Therefore, we do not want to make any changes to vehicle_model for scenario_simulator_v2.
Please fix the issue without modifying vehicle_model.

yamacir-kit (changes requested):

simulation/simple_sensor_simulator/include/simple_sensor_simulator/vehicle_simulation/vehicle_model/sim_model_delay_steer_acc_geared.hpp
dmoszynski: @yamacir-kit 
Please take a look at this part of the description:

> ### [Oz position in SimModelInterface](https://github.com/tier4/scenario_simulator_v2/pull/1226/commits/46490d6dd7239416c39905652878692b19dee53b)
> - This change is necessary for FollowTrajectoryAction to work properly. Entities move along lanelets, where pitch and Oz are taken into account. If FollowTrajectoryAction is to overwrite Ego positions, the position in Oz cannot just be ignored.
> - Currently, the virtual `getZ()` method has been implemented, which returns 0 for every model except the one currently in use.
> - The position in Oz is set only when the position is overwritten, it is not calculated in any way. To be more precise, if the `update()` in SimModelInterface is used, everything still works on 2D, and the position in Oz is only used during `overwrite()` - that is, only by FollowTrajectoryAction.
> - I believe that all SimModelInterface should be adapted to 3D, due to the fact that all the rest of the code in ss2 currently uses 3D and Autoware uses 3D.

The issue that occurs is related to the lack of consideration of the `Oz` axis in the calculations. As a result, after each update the entire position (in the `Ox` and `Oy` axes also) is inconsistent with that which was overwritten. 

As far as I understand, on the `Autoware` side both `Oz` and pitch are taken into account. Recently, complete consideration of both `Oz` and pitch was introduced on the `scenario_simulator_v2` side. 

The issue I encountered is related to the fact that in the [SimModelInterface](https://github.com/tier4/scenario_simulator_v2/pull/1226/commits/46490d6dd7239416c39905652878692b19dee53b) axis `Oz` is not taken into account. This causes that during communication `EgoEntitySimulation->SimModelInteface->Autoware` there is a conversion between `3D->2D->3D` which introduces inaccuracies (position differences), and a simple alignment to `lanelet` (matching `Oz` position to `lanelet`) does not solve the issue, because the values in `Ox` and `Oy` are also affected and not the same. 

Therefore, the only alternative solution I can propose for the proper operation of `FollowTrajectotryAction` for Ego (when `isControlledBySimulator()==true`) is to completely ignore the position received from `Autoware`. This implies that the status in `EgoEntitySimulation` will not be calculated using `SimModelInterface` - it will simply remain overwritten without “getting the values from Autoware”.

However, I believe that it is inevitable to introduce the `Oz` axis in `SimModelInteface` to make the positions passed between `scenario_simulator_v2` and `Autoware` the same.

yamacir-kit: @dmoszynski 
I understand that this change is essential due to the request for accuracy.

However, since simple_planning_simulator is still the one most often used by Autoware's planner development team, its vehicle_model and ssv2's vehicle_model should be kept equivalent.

Therefore, I will now propose to simple_planning_simulator to change about `getZ` in this pull request. After it is confirmed that there is no problem with the proposal, I will ask you to send a pull request to autowarefoundation/autoware. I will then justify the changes in this pull request by saying “import the changes in autowarefoundation/autoware”.

Thanks for your detailed explanation.

dmoszynski: > I will now propose to simple_planning_simulator to change about getZ in this pull request. After it is confirmed that there is no problem with the proposal, I will ask you to send a pull request to autowarefoundation/autoware. 

Thank you so much, then I'm waiting for information about the ability to create a PR.

yamacir-kit: @dmoszynski 
I am revisiting the changes required by this pull request in order to propose changes to the planner development team. So, I am wondering about the following:

- IDX::Z is not involved in the calculations inside SimModel.
- IDX::Z exists only to transmit the value of the Z coordinate from EgoEntitySimulation::overwrite to EgoEntitySimulation::getCurrentPose.
- Therefore, IDX::Z does not need to be included in SimModel and can be replaced by a data member `z` in the EgoEntitySimulation class.

Please let me know if the above is correct. If it is correct, there is no need to change simple_planning_simulator, so please change the code as above.

dmoszynski: @yamacir-kit 
Thank you for pointing this out, I developed this code more than a month ago, I will take a closer look at it and let you know why I concluded ( and if right) that such changes are required.

dmoszynski: @yamacir-kit 
I've developed a [solution with no need to modify](https://github.com/tier4/scenario_simulator_v2/pull/1226/commits/6024371b9b0845cfc3ca7e47e6874b9de00ef5fb) `SimModelInterface`. I've performed tests using the scenarios from the PR description - it confirms the solution to the issue - videos below. I run evaluator for the current version of the branch, I will share the results on the jira ticket. 
#### Scenario with slopes - EGO

https://github.com/tier4/scenario_simulator_v2/assets/121798334/23fc7b5c-5292-41a4-a7ce-bfba5289d99a


#### [**autoware.yaml](https://github.com/tier4/scenario_simulator_v2/blob/master/test_runner/scenario_test_runner/scenario/RoutingAction.FollowTrajectoryAction-autoware.yaml) - EGO

https://github.com/tier4/scenario_simulator_v2/assets/121798334/9949dd87-e76f-4049-9ae5-0dc04c7b1a9a

dmoszynski:
@yamacir-kit 
PR is ready for re-review ([regressions results](https://tier4.atlassian.net/browse/RJD-995?focusedCommentId=170929)).

simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp
```diff
@@ -419,26 +440,20 @@ auto EgoEntitySimulation::getCurrentTwist() const -> geometry_msgs::msg::Twist
...
-    geometry_msgs::msg::Vector3 rpy;
-    rpy.x = 0;
-    rpy.y = pitch_angle;
-    rpy.z = vehicle_model_ptr_->getYaw();
-    return initial_pose_.orientation * quaternion_operation::convertEulerAngleToQuaternion(rpy);
-  }();
-
-  return current_pose;
+  const auto relative_position =
+    quaternion_operation::getRotationMatrix(initial_pose_.orientation) * world_relative_position_;
```
MertClk, line 444: https://eigen.tuxfamily.org/dox/TopicPitfalls.html

- https://github.com/tier4/scenario_simulator_v2/issues/1302

```cpp
const Eigen::Vector3d relative_position =
```

This should fix it.

hakuturu583: Thank you for finding problem!
I also think it should be fix.

MertClk: Thank you @hakuturu583, it was my pleasure.

Issue #1193, opened 2024-02-16
simple_sensor_simulator outputs shows any objects in occulusion


HansRobo:
**Describe the bug**

In the two images below, a vehicle exists in a blind spot caused by the MiscObject in the foreground.
But, We can see the vehicle in both of perception output and gridmap output from simple_sensor_simulator. 

![image](https://github.com/tier4/scenario_simulator_v2/assets/12395284/029361ec-9f61-4787-a81f-5c2da6a12ac2)
![image](https://github.com/tier4/scenario_simulator_v2/assets/12395284/a99641ed-75e1-439c-8902-6dc75572c90b)

**To Reproduce**

- spawn vehicle in occulusion spot
- check perception and gridmap outputs from `simple_sensor_simualtor`

**Expected behavior**

Entities in blind spots are not output to topics for Autoware.

**Screenshots**
I put them in "Describe the bug" section.

**Desktop (please complete the following information):**
 - Ubuntu Version: 22.04
 - ROS 2 version: Humble
 - DDS: cyclone DDS

PR #1227, opened 2024-04-11
Exclude LiDAR occluded object on OccupancyGrid

f0reachARR:
# Description
## Abstract
Partial fixes for #1193 

## Background

OccupancyGridSensor in simple_sensor_simulator takes detected objects from LidarSensor, but never uses it when `filter_by_range` is true.
This causes occluded objects (objects that were not detected) appeared in OccupancyGrid (explained in #1193).

## Details
Filter objects using detected objects in `getDetectedObjects` method.

Before
image

After
image

As you see pictures, occluded object (tom, orange) emitted before this PR.

# Reference

https://github.com/tier4/sim_evaluation_tools/issues/270

# Destructive Changes

OccupancyGridSensor behavior was changed.
It emits only LiDAR detected (non occluded) objects instead of all objects in sensor range.

# Known Limitations

As LiDAR origin passed to Raycaster is too low (Ego bottom), detected objects from LidarSensor contain some objects that should be occluded.
This can be observed in DetectionSensor output in #1193


hakuturu583 (changes requested):
Code looks good, but this pull request will changes the sensor output.
I think it is destructive change, so please describe it with sample image in pull request description.

f0reachARR:
I added two images to PR description. Thank you.

hakuturu583 (approved):

PR #1238, opened 2024-05-01
Publish `entities` frame

hakuturu583:
# Description

## Abstract

When I checking performance of scenario_simulator_v2, I found that it takes a lot of time for broadcasting frames for each entity.

## Background

When spawning a lot of entity (About 10), it takes a lot of times to broadcasting frames.
It causes frame drop, so we have to fix it.

## Details

Broadcast `entities` frame, which is based on the average position of all Entities.
This PR is depends on #1237 

## References

#1237 
https://github.com/tier4/sim_evaluation_tools/issues/274

# Destructive Changes

In previous version, all of the entity has tf frame and it was used in visualization, but it is removed in this pull request.

# Known Limitations

N/A

### DIFF
--- a/openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
+++ b/openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
@@ -162,7 +162,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
   auto evaluate_storyboard = [this]() {
     withExceptionHandler(
       [this](auto &&...) {
-        publishCurrentContext();
+        // publishCurrentContext();
         deactivate();
       },
       [this]() {
@@ -183,7 +183,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
 
           SimulatorCore::update();
 
-          publishCurrentContext();
+          // publishCurrentContext();
         });
       });
   };
@@ -193,7 +193,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
   } else {
     return withExceptionHandler(
       [this](auto &&...) {
-        publishCurrentContext();
+        // publishCurrentContext();
         reset();
         return Interpreter::Result::FAILURE;  // => Inactive
       },
--- a/openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
+++ b/openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
@@ -162,7 +162,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
   auto evaluate_storyboard = [this]() {
     withExceptionHandler(
       [this](auto &&...) {
-        // publishCurrentContext();
+        publishCurrentContext();
         deactivate();
       },
       [this]() {
@@ -183,7 +183,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
 
           SimulatorCore::update();
 
-          // publishCurrentContext();
+          publishCurrentContext();
         });
       });
   };
@@ -193,7 +193,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
   } else {
     return withExceptionHandler(
       [this](auto &&...) {
-        // publishCurrentContext();
+        publishCurrentContext();
         reset();
         return Interpreter::Result::FAILURE;  // => Inactive
       },
--- a/simulation/traffic_simulator/src/entity/entity_manager.cpp
+++ b/simulation/traffic_simulator/src/entity/entity_manager.cpp
@@ -16,6 +16,7 @@
 #include <geometry/bounding_box.hpp>
 #include <geometry/distance.hpp>
 #include <geometry/intersection/collision.hpp>
+#include <geometry/linear_algebra.hpp>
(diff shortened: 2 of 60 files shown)


simulation/traffic_simulator/src/entity/entity_manager.cpp
```diff
@@ -38,12 +39,27 @@ namespace entity
 void EntityManager::broadcastEntityTransform()
 {
   std::vector<std::string> names = getEntityNames();
-  for (const auto & name : names) {
-    geometry_msgs::msg::PoseStamped pose;
-    pose.pose = getMapPose(name);
-    pose.header.stamp = clock_ptr_->now();
-    pose.header.frame_id = name;
-    broadcastTransform(pose);
+  if (entityExists("ego")) {
```
HansRobo, line 42: Please do not use "ego" directly and use `getEgoName` instead.

HansRobo: But this is just a nit, and does not block merging.

simulation/traffic_simulator/src/entity/entity_manager.cpp
```diff
@@ -38,12 +39,27 @@ namespace entity
...
+      geometry_msgs::build<geometry_msgs::msg::PoseStamped>()
+        .header(
+          std_msgs::build<std_msgs::msg::Header>().stamp(clock_ptr_->now()).frame_id("entities"))
+        .pose(geometry_msgs::build<geometry_msgs::msg::Pose>()
+                .position(std::accumulate(
+                  names.begin(), names.end(), geometry_msgs::msg::Point(),
+                  [this, names](geometry_msgs::msg::Point & point, const std::string & name) {
+                    return point +
+                           (getMapPose(name).position * (1.0 / static_cast<double>(names.size())));
+                  }))
```
HansRobo, line 60: What is this calculation for?

hakuturu583: This part of the process is intended to ensure that frames are issued in a position that makes it as easy as possible to see the entities that will appear in the scenario.

hakuturu583: In the past, we used to publish the frames of all entities, but that would be too heavy processing, so we publish the average of the coordinates of all entities.

simulation/traffic_simulator/src/entity/entity_manager.cpp
```diff
@@ -38,12 +39,30 @@ namespace entity
...
-    pose.pose = getMapPose(name);
-    pose.header.stamp = clock_ptr_->now();
-    pose.header.frame_id = name;
-    broadcastTransform(pose);
+  if (isEgoSpawned()) {
+    const auto ego_name = getEgoName();
+    if (entityExists(ego_name)) {
+      broadcastTransform(
+        geometry_msgs::build<geometry_msgs::msg::PoseStamped>()
+          .header(std_msgs::build<std_msgs::msg::Header>().stamp(clock_ptr_->now()).frame_id("ego"))
```
HansRobo, line 47:
```suggestion
          .header(std_msgs::build<std_msgs::msg::Header>().stamp(clock_ptr_->now()).frame_id(ego_name))
```

openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
```diff
@@ -162,7 +162,7 @@ auto Interpreter::on_activate(const rclcpp_lifecycle::State &) -> Result
   auto evaluate_storyboard = [this]() {
     withExceptionHandler(
       [this](auto &&...) {
-        publishCurrentContext();
+        // publishCurrentContext();
```
HansRobo, line 165: Why this line is commented out?

hakuturu583: Sorry, I fixed this problem.

simulation/traffic_simulator/src/entity/entity_manager.cpp
hakuturu583: This is the intended implementation.
It is easier to create rviz config if the name “ego” is fixed, so the frame_id “ego” is issued regardless of the name of the ego entity.

HansRobo: I see, I understand.

simulation/traffic_simulator/src/entity/entity_manager.cpp
HansRobo: Clearly documented by https://github.com/tier4/scenario_simulator_v2/pull/1238/commits/0fe90566ed68e5e13710eec37bbf6ac769cda0ee

simulation/traffic_simulator/src/entity/entity_manager.cpp
HansRobo: Fixed by https://github.com/tier4/scenario_simulator_v2/pull/1238/commits/61af6463271fb24640a32fbb075e921d31677cb0

simulation/traffic_simulator/src/entity/entity_manager.cpp
HansRobo: I checked well-documented in https://github.com/tier4/scenario_simulator_v2/pull/1238/commits/9b5444900e2ac529131c5bf1ab5d4481465baf49

HansRobo (approved):

PR #1248, opened 2024-05-10
add `if` section in the restore branch step of the Release action

hakuturu583:
# Description

## Abstract

Fixed a bug that caused release actions to fail when an external contributor opened a pull request from a forked repository.

## Background

This step is intended to restore a branch that was inadvertently deleted at the same time as a merge, and is set up so that it can work even if there is an operational error.
However, when a pull request is received from an external repository, the branch to be restored does not exist and an error occurs.

## Details

I introduced an option to ignore the error so that pull requests from external repositories can be merged.

## References

#1242 

# Destructive Changes

N/A

# Known Limitations

N/A


HansRobo:
I would appreciate it if you could work on this PR in your spare time.
(Every morning, I get a notification from GitHub to review this.)

PR #1266, opened 2024-05-28
ref(traffic_simulator): extend utils/pose - use it globally, improve canonization process

dmoszynski:
# Description

This PR is based on [TMP_PR1](https://github.com/tier4/scenario_simulator_v2/pull/1251) - for which regression tests have been run and no regression has been [evidenced](https://tier4.atlassian.net/browse/RJD-1053?focusedCommentId=170334).

## Abstract
This PR is intended to separate methods related to pose calculation from `API`/`EntityManager`/`EntityBase` modules. 

In addition, PR improves the canonicalization of `msg::LaneletPose` and guarantees that the matching of `msg::Pose` to lanelet takes place in the `CanonicalizedEntityStatus` constructor - as long as `CanonicalizeLaneletPose` exists, i.e. `msg::LaneletPose` is valid.

Methods that have been completely separated in this PR:
- `canonicalize()`
- `toMapPose()`
- `toLaneletPose()`
- `getMapPoseFromRelativePose()` (it is not getter)
- `getLaneletLength()` (it is not getter)

The result is a collection o free functions in namespace [traffic_simulator::pose](https://github.com/tier4/scenario_simulator_v2/blob/ref/RJD-1053-implement-pose-utils/simulation/traffic_simulator/include/traffic_simulator/utils/pose.hpp).

Methods (getters) whose forwards have been removed:
- `getMapPose()`
- `getLaneletPose()`

**Note: This PR modifies a very large number of files and lines, reason for this is the separation of API::canonicalized() to pose::canonicalized() and API::toMapPose() to pose::toMapPose() - which are [used very commonly](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-d4b6a32da8891f6e24d140536cf07919a8a5d5eb00b4c05363fbdc7f6e78acac).**

**Note:  The continuation of this PR is [another PR](https://github.com/tier4/scenario_simulator_v2/pull/1252/files) that improves and unifies responsibility for canonicalize - please take this into account when reviewing this PR.**

## Background
Now the algorithms used for pose are implemented directly in `EntityManager` and `EntityBase`. This makes it necessary to call forwarded methods to perform any pose conversions in `SimulatorCore`. The new approach assumes that `SimulatorCore` will have access to each Entity (`EntityBase`) and lanelet2 (`HdmapUtils`) and will perform conversions using an independent library. This ensures that the algorithms for the calculations are separated and their testing will be much more efficient. Moreover, it definitely reduces the complexity of `EntityManager` and `EntityBase`.

Another globally occurring problem is the non-distinction between `LaneletPose` and `CanonicalizedLaneletPose`. What is more, `canonicalize()` is performed in various parts of the code in different ways (not always arguments with the right value are used). In addition, in some cases there is a need to adjust the position in `Oz` axis to lanelet and `pitch` orientation. This match should be performed with each canonicalization but now it is not - this also introduces this PR.

#### 1.1. Distinction between `LaneletPose` and `CanonicalizedLaneletPose`
- Ensured that `getLaneletPose()` returns `LaneletPose` but [`getCanonicalizedLaneletPose()`](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-c9369b365c55e08a9f09306243a8b4be81d4feffb41ee8cfff73890136326d42) returns `CanonicalizedLaneletPose`
- Developed `toCanonicalizedLaneletPose()`, which returns `CanonicalizedLaneletPose` and ensured that it is used globally.

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/api/api.cpp#L132-L136

#### 1.2. Matching msg::Pose to lanelet
- `adjustOrientationAndOzPosition()` has been added to the `CanonicalizedLaneletPose` costructor, which provides a match of `map_pose_` to the lanelet - more specifically, the position in Oz and the pitch orientation. 

Note: Result of ths match is used during canonicalization of the `EntityStatus` (in the `CanonicalizedEntityStatus` constructor)

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/data_type/lanelet_pose.cpp#L132-L157

#### 1.3. consider_pose_by_road_slope
- Added a static variable `consider_pose_by_road_slope_`, which is used when the position is matching relative to the lanelet
- Added declaration of  `consider_pose_by_road_slope` ROS parameter in the nodes where it was missing, and ensured that it is loaded and set using `CanonicalizedLaneletPose::setConsiderPoseByRoadSlope()` (cpp_mock_scenarios, open_scenario_interpreter, simple_sensors_simulator, random_test_runner).

Note: Earlier this parameter was not always used, now the risk of not taking it into account has been minimized.

Note: It might be worth to add a check if for a specific node `CanonicalizedLaneletPose::setConsiderPoseByRoadSlope()` was called before the first use of the `consider_pose_by_road_slope_` variable, if not then throw an exception.

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/include/traffic_simulator/data_type/lanelet_pose.hpp#L84

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/mock/cpp_mock_scenarios/launch/mock_test.launch.py#L145-L149

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/mock/cpp_mock_scenarios/src/cpp_scenario_node.cpp#L34-L41

#### 1.4 Canonization process
- Now wherever the pose needs to be matched to the lanelet `toCanonicalizedLaneletPose()` is called, the returned `CanonicalizedLaneletPose` is passed to the `CanonicalizedEntityStatus()` constructor to guarantee that it has the right canonicalization of `msg::LaneletPose` and the match of `msg::Pose` to the lanelet. 

Note: Canonicalization needs specific arguments that are not always available (like `matching_distance` for a specific entity). Moving the canonicalization process to one part of the code will solve this problem - this will be introduced in the [next PR](https://github.com/tier4/scenario_simulator_v2/pull/1252).

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L473-L487

#### 1.5 Helper extension
- [Added](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-93a103269043a7c2f66f7ce97436c40606603492be050ef1e453932b0eed3eb2) `helper::contructCanonicalizedLaneletPose()`, which is used mostly in cpp_mock_scenarios

### 2. CanonicalizedEntityStatus
- Added `canonicalized_lanelet_pose_` field type `CanonicalizedLaneletPose`
  - this minimizes the number of operations when calling `getCanonicalizedLaneletPose()`
  - guarantees that always the returned `msg::LaneletPose` is right canonicalized
  - guarantees that always returned `msg::Pose` is matched to the lanelet (if valid)
- Added constructor `CanonicalziedEntityStatus(EntityStatus, CanonicalizedLaneletPose)` which allows to avoid additional canonicalization and increases probability of its correct result (canonicalization inside `CanonicalizedEntityStatus` doesn't have access to `matching_distance`). 
- Added methods `getLaneletId()` and `getLaneletIds()` to [improve code readability](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-970ca1021b8bc8b7ca7790a53aefc7ad6e49d457a56aa7b261cbfd53b4f89af9)

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/data_type/entity_status.cpp#L23-L30

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/data_type/entity_status.cpp#L68-L89

### 3. EgoEntitySimulation
- [Removed](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-a70b23781f2bdd4b04e18eb52c51dd37fd686738ec2b720c1ec3c5fc6844e462) `getMatchedLaneletPoseFromEntityStatus()` and used `toCanonicalizedLaneletPose()` instead.
- Removed `fillLaneletDataAndSnapZToLanelet()`, since the matching that was done there now takes place in the `CanonicalizedLaneletPose()` constructor.
- Removed `setInitialStatus()` and ensured that `init_status` is passed as an argument to the constructor - it also calls `setStatus()` which ensures canonicalization of the initialization pose

Note: Now inside `setStatus()`, `toCanonicalizedLaneletPose()` is called, and the result is passed to the `CanonicalizedEntityStatus()` constructor, which is assigned to `status_` - which stores the current state of Ego. 

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp#L448-L462

### 4. API
- Ensured that `spawn()` accepts pose as `LaneletPose`
- Improved `setEntityStatus()` to be uniform and always use `toCanonicalizeLaneletPose()`, the result `canonicalized_lanelet_pose` is passed to the `CanonicalizedEntityStatus()` constructor

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/api/api.cpp#L101-L141

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/src/api/api.cpp#L160-L209

### 5. EntityManager
- `spawnEntity()` has been significantly simplified and improved, highlighted how each of the 3 possible types is handled: `msg::Pose`, `msg::LaneletPose`, `CanonicalizedLaneletPose`  - `toCanonicalizeLaneletPose()` with `bounding_box` and `matching_distance` passed takes place only in the case of `msg::Pose`
- Simplified `isInLanelet()` - separate `pose::isInlanelet()`, which is used

https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L455-L488

### 6. Others
- Added `pose::isAtEndOfLanelets()` and used it in `PedestrianEntity::onUpdate()` and `VehicleEntity::onUpdate()`
- `fillLaneletPose()` has been removed from every `EntityBase`, `PedestrianEntity`, `VehicleEntity`, `EgoEntity`.
- [Separated](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-1b86cbb79a60ba1de3f7e2b1918ee48d4f7b1501f7d1568b938f261ec3bae671) `estimateLaneletPose()` from `PedestrianActionNode` as `pose::estimateCanonicalizedLaneletPose()`
- Modified the `ByEntityCondition.EntityCondition.RelativeDistanceConditionFreespace.yaml` scenario, as the introduction of improvements in canonicalization and matching positions to the lanelet for each `CanonicalizedEntityStatus()` caused very [small differences in positions (on the order of **1e-11**)](https://github.com/tier4/scenario_simulator_v2/pull/1266/files#diff-82f48796712ea4e5a31569cd0976552e9ca24c8ef547ed27f3a4914adb166959).

#### API
Some of the `API::` methods have been removed, free functions should be used instead.

Previously, these were forwarded functions to the entity manager, which used the object of the specific entity. Now it is necessary to get the selected entity using `API::getEntity()`, extract the required objects from it and then pass them to freefunctions of the `pose::` collection. If `lanelet2` is required for the calculation, it is required to pass `HdMapUtils` as `shared_ptr` - always as the last argument.  

Details are shown in the table.
All functions in the following tables require the `traffic_simulator` namespace. (e.g `API::getMapPose` is `traffic_simulator::API::getMapPose`)

<table>
  <thead>
    <tr>
      <th>Legacy function</th>
      <th>New function</th>
      <th>Example of usage</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td><code>API::canonicalize</code></td>
      <td><code>pose::canonicalize</code></td>
       <td><a href="https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L482">Link</a></td>
    </tr>
    <tr>
      <td><code>API::toMapPose</code></td>
      <td><code>pose::toMapPose</code></td>
       <td><a href="https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/mock/cpp_mock_scenarios/src/follow_lane/acquire_position_in_world_frame.cpp#L57-L59">Link</a></td>
    </tr>
    <tr>
      <td><code>API::toLaneletPose</code></td>
      <td><code>pose::toCanonicalizedLaneletPose</code></td>
       <td><a href="https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp#L457-L459">Link</a></td>
    </tr>
    <tr>
      <td><code>API::getMapPoseFromRelativePose</code></td>
      <td><code>pose::transformRelativePoseToGlobal</code></td>
       <td><a href="https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/mock/cpp_mock_scenarios/src/random_scenario/random001.cpp#L228-L232">Link</a></td>
    </tr>
    <tr>
      <td><code>API::getLaneletLength</code></td>
      <td><code>pose::laneletLength</code></td>
       <td><a href="https://github.com/tier4/scenario_simulator_v2/blob/186c8309c9cd72aa5d4cba344f466c2dd7c50e99/mock/cpp_mock_scenarios/src/random_scenario/random001.cpp#L136">Link</a></td>
    </tr>
  </tbody>
</table>

#### CppScenarioNode
CppScenarioNode::spawnEgoEntity() method has changed, from 

https://github.com/tier4/scenario_simulator_v2/blob/45d42a79d92c370387749ad16c10665deb42e02c/mock/cpp_mock_scenarios/include/cpp_mock_scenarios/cpp_scenario_node.hpp#L50-L53

to 

https://github.com/tier4/scenario_simulator_v2/blob/cac70ef1c2846d6bd1dba477366befb14b323341/mock/cpp_mock_scenarios/include/cpp_mock_scenarios/cpp_scenario_node.hpp#L50-L53

This has been introduced to simplify spawning in `cpp_mock` scenarios, now there is no need to canonicalize `LaneletPose` first.

Example:

https://github.com/tier4/scenario_simulator_v2/blob/cac70ef1c2846d6bd1dba477366befb14b323341/mock/cpp_mock_scenarios/src/respawn_ego/respawn_ego.cpp#L56-L59


yamacir-kit (changes requested):

hakuturu583:
@dmoszynski Please adding migration guide about `void CppScenarioNode::spawnEgoEntity();`.
This function is used by almost all of scenarios written in C++.

simulation/traffic_simulator/src/utils/pose.cpp
```diff
@@ -45,18 +46,31 @@ auto quietNaNLaneletPose() -> traffic_simulator::LaneletPose
...
-auto toLaneletPose(
-  const geometry_msgs::msg::Pose & map_pose, bool include_crosswalk,
+auto toMapPose(
+  const LaneletPose & lanelet_pose,
+  const std::shared_ptr<hdmap_utils::HdMapUtils> & hdmap_utils_ptr) -> geometry_msgs::msg::Pose
+{
+  return hdmap_utils_ptr
+    ->toMapPose(lanelet_pose, CanonicalizedLaneletPose::getConsiderPoseByRoadSlope())
+    .pose;
+}
```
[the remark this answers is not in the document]
dmoszynski: @yamacir-kit 
By “has been completely separated” I mean that now everywhere in traffic_simulator `pose::toMapPose` is used - there is no use of `HdMapUtils::toMapPose` directly anywhere. Namespace `pose::` has free functions that use `HdMapUtils`, some of them are more complex, others are simple calls to function from `HdMapUtils`. 

I've intentionally not moved the implementation from `HdMapUtils::toMapPose` to `pose::toMapPose` at this time, as the goal of this PR is not to refactor `HdMapUtils`. During `HdMapUtils` refactoring (which [I am also in charge of](https://tier4.atlassian.net/browse/RJD-1089)), I would like to split the architecture into `utils` and `lanelet_core` (as in the attached draft image). 
utils drawio


`HdMapUtils::toMapPose` will probably be moved to `lanelet_core::pose`, since the calculations using `msg::LaneletPose` and `HdMapUtils::getCenterPointsSpline()` are performed there - so it's “near lanelet”

simulation/traffic_simulator/src/data_type/entity_status.cpp
```diff
@@ -19,56 +19,78 @@ namespace traffic_simulator
...
+  assert(entity_status_.lanelet_pose_valid == canonicalized_lanelet_pose_.has_value());
+  if (canonicalized_lanelet_pose_) {
+    entity_status_.lanelet_pose_valid = true;
+    entity_status_.lanelet_pose = static_cast<LaneletPose>(canonicalized_lanelet_pose_.value());
+    /*
+      The Oz position and orientation based on LaneletPose are rewritten to the used map_pose,
+      since such adjustment relative to the lanelet is necessary,
+      The position in Ox and Oy is not rewritten because the map_pose retrieved via
+      lanelet_pose = toCanonicalizedLaneletPose(map_pose), then map_pose = toMapPose(lanelet_pose)
+      can be different from the original one (especially if the entity changes lane)
```
hakuturu583, line 79: What is `Ox` ,`Oy`, `Oz` ?

simulation/traffic_simulator/src/utils/pose.cpp
```diff
@@ -160,5 +223,111 @@ auto boundingBoxRelativeLaneletPose(
...
+    const auto start_edge =
+      helper::constructCanonicalizedLaneletPose(lanelet_id, 0.0, 0.0, hdmap_utils_ptr);
+    const auto end_edge = helper::constructCanonicalizedLaneletPose(
+      lanelet_id, hdmap_utils_ptr->getLaneletLength(lanelet_id), 0.0, hdmap_utils_ptr);
+    auto dist0 = longitudinalDistance(
+      start_edge, canonicalized_lanelet_pose, include_adjacent_lanelet, include_opposite_direction,
+      allow_lane_change, hdmap_utils_ptr);
+    auto dist1 = longitudinalDistance(
+      canonicalized_lanelet_pose, end_edge, include_adjacent_lanelet, include_opposite_direction,
+      allow_lane_change, hdmap_utils_ptr);
```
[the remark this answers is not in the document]
dmoszynski: Done

simulation/traffic_simulator/src/utils/pose.cpp
```diff
@@ -160,5 +223,111 @@ auto boundingBoxRelativeLaneletPose(
...
+  constexpr bool include_opposite_direction{false};
+  constexpr bool allow_lane_change{false};
+
+  if (isSameLaneletId(canonicalized_lanelet_pose, lanelet_id)) {
+    return true;
+  } else {
+    const auto start_edge =
+      helper::constructCanonicalizedLaneletPose(lanelet_id, 0.0, 0.0, hdmap_utils_ptr);
+    const auto end_edge = helper::constructCanonicalizedLaneletPose(
+      lanelet_id, hdmap_utils_ptr->getLaneletLength(lanelet_id), 0.0, hdmap_utils_ptr);
```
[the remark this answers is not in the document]
dmoszynski: Done

simulation/behavior_tree_plugin/src/action_node.cpp
```diff
@@ -408,6 +407,7 @@ auto ActionNode::calculateUpdatedEntityStatus(
   geometry_msgs::msg::Accel accel_new = std::get<1>(dynamics);
   geometry_msgs::msg::Twist twist_new = std::get<0>(dynamics);
 
+  /// @note here is no entity_status->laneMatchingSucceed() check -> it may throw an exception
```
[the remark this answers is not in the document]
dmoszynski: Done

hakuturu583:
> @dmoszynski Please adding migration guide about `void CppScenarioNode::spawnEgoEntity();`. This function is used by almost all of scenarios written in C++.

Thanks, I confirmed that descriptions are in the migration guide.

simulation/traffic_simulator/src/data_type/entity_status.cpp
hakuturu583: It is a little clearer, but the definition of Ox/Oy/Oz is unclear.
Please put “X/Y/Z values of the Cartesian coordinate system with respect to map” in the comment instead of omitting them.

When writing comments, please try as much as possible to write comments that can be interpreted correctly even without prior knowledge of the meaning.

hakuturu583 (changes requested):

hakuturu583:
I believe that [this document](https://github.com/tier4/scenario_simulator_v2/tree/master/docs/developer_guide/lane_pose_calculation) may need to be revised in accordance with this PR. This is a document that we have been working on for a long time.

Please check and correct if necessary.

simulation/traffic_simulator/src/utils/pose.cpp
```diff
@@ -160,5 +224,114 @@ auto boundingBoxRelativeLaneletPose(
...
+/*
+  This function has been moved from pedestrian_action_node and modified,
+  in case of inconsistency please compare in original:
+  https://github.com/tier4/scenario_simulator_v2/blob/090a8d08bcb065d293a530cf641a953edf311f9f/simulation/behavior_tree_plugin/src/pedestrian/pedestrian_action_node.cpp#L67-L128
+*/
+auto estimateCanonicalizedLaneletPose(
+  const geometry_msgs::msg::Pose & map_pose,
+  const traffic_simulator_msgs::msg::BoundingBox & bounding_box,
+  const lanelet::Ids & unique_route_lanelets, const bool include_crosswalk,
+  const double matching_distance, const std::shared_ptr<hdmap_utils::HdMapUtils> & hdmap_utils_ptr)
```
hakuturu583, line 237: The PedestrianActionNode::estimateLaneletPose function could only be used for PedestrianEntity, but making it a free function would create the possibility of misuse.
Please emphasize that this function should only be applied to the pedestrian entity by adding the pedestrian namespace

hakuturu583 (changes requested):

simulation/traffic_simulator/src/utils/pose.cpp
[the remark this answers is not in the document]
dmoszynski: https://tier4.atlassian.net/browse/RJD-1089?focusedCommentId=172725

dmoszynski:
> I think that [this document](https://github.com/tier4/scenario_simulator_v2/tree/master/docs/developer_guide/lane_pose_calculation) may need to be revised in accordance with this PR. This is a document that we have been working on for a long time.
> 
> Please check and correct if necessary.

@hakuturu583 

Thank you so much for your valuable recommendation. I didn't pay attention to it earlier. 
I've revised the [documentation](https://github.com/tier4/scenario_simulator_v2/tree/ref/RJD-1053-set-update-canonicalized-entity-status/docs/developer_guide/lane_pose_calculation), but decided to introduce it in the [next PR](https://github.com/tier4/scenario_simulator_v2/pull/1252/files) for this task - because it also introduces significant changes with lane pose calculations. (target PR to master will be created after this PR is merged).

simulation/traffic_simulator/src/utils/pose.cpp
dmoszynski: @hakuturu583 
I believe that introducing namespace related to entity type into `utils/pose` may result in code complexity - it will decrease its readability.

- I understand your concern, and indeed it increases the probability that `estimateCanonicalizedLaneletPose` will be used for a non-Pedestrian entity. 
- However, I think it would be a better idea to change the name to suggest that the method has an additional match rather than the classic “toCanonicalizedLaneletPose”, if “estimate” itself does not suggest it. Or just add a more informative comment.
- It is unusual here to set the position “at the end of the lanelet”. - maybe some name related to this? 

Meanwhile, I believe the method is generic enough to be worth keeping it in `utils/pose` instead of directly in `PedestrianActionNode`.

hakuturu583: > Meanwhile, I believe the method is generic enough to be worth keeping it in utils/pose instead of directly in PedestrianActionNode.

Yes. I agree.
I want to talk about the better name of the function `estimateCanonicalizedLaneletPose`

hakuturu583: How about 

- pedestrian::estimateLaneletPose
- pedestrian::computeLaneletPose
- pedestrian::transformToLaneletPose
- pedestrian::matchToLaneletPose

hakuturu583: I think `pedestrian::transformToLaneletPose` is best.

simulation/traffic_simulator/src/utils/pose.cpp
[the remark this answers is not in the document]
yamacir-kit: @dmoszynski 
I understand. I mark this conversation as resolved.

simulation/traffic_simulator/src/utils/pose.cpp
dmoszynski: @hakuturu583 
I thought of a name like `traffic_simulator::pose::toCanonicalizeLaneletPoseOrFallback`, since the `estimateCanonicalizedLaneletPose` method truly use `toCanonicalizedLaneletPose`, and if that fails, uses a fallback option - it uses hardcoded matching_distance == 2, and if that also fails, it uses `LaneletPose` at the end of the. lanelet. 

What do you think about it?

hakuturu583: I think `toCanonicalizeLaneletPoseOrFallback` is not a good naming because of we could not guess `Fallback to what` and `Fall back from waht` from the function's name.

dmoszynski: By fallback I just mean “a fallback solution when `toCanonicalizedLanePose` fails” - that is, using hardcoded `matching_distance`, and if that fails using the position at the end of the lane. “`toCanonicalizedLaneletPoseOrFallback`” tells us that `toCanonicalizedLaneletPose()` is called in the function, and if it fails fallback solutions are used. Maybe a better choice instead of “fallback” is to use “contingency” here.

I thought of two more proposals:
`pose::toCanonicalizeLaneletPoseKeepingLanelet` or `pose::toCanonicalizeLaneletPoseKeepingRoute`.

Naturally, if the names proposed by me are confusing, I understand and use the name `pedestrian::transformToLaneletPose` (more precisely `pedestrian::transformToCanonicalizedLaneletPose)`, as you proposed. :bow:

hakuturu583: I am a developer of this software, so I can guess `KeepingLanelet` and `KeepingRoute` means.
But if I am not a developer of this software, I think I could not guess it's meanings.
Please use `pedestrian::transformToCanonicalizedLaneletPose`

dmoszynski: @hakuturu583 
Of course, [done](https://github.com/tier4/scenario_simulator_v2/pull/1266/commits/6f87603e2aaaddfd8ceb08645cb3933fe3b74515#diff-1b86cbb79a60ba1de3f7e2b1918ee48d4f7b1501f7d1568b938f261ec3bae671) :bow:

hakuturu583 (approved):

hakuturu583 (commented):
Discussing with maintainers.
Please wait.

hakuturu583 (changes requested):
Discussing with maintainers.
Please wait.

dmoszynski:
@hakuturu583 
Of course, I'm waiting for information. 
I would also like to propose to run the regression tests again, as the changes made during the review were not subtle.
I ran local tests and they were successful, but I think it is worth to check with the evaluator.
(so far, for this branch, regressions tests have been run only once - before the tier4 review)

yamacir-kit (changes requested):
Minor change requests. These change requests do not require another regression test with the Autoware Evaluator.

mock/cpp_mock_scenarios/include/cpp_mock_scenarios/cpp_scenario_node.hpp
```diff
@@ -48,7 +48,7 @@ class CppScenarioNode : public rclcpp::Node
     return false;
   }
   void spawnEgoEntity(
-    const traffic_simulator::CanonicalizedLaneletPose & spawn_lanelet_pose,
+    const traffic_simulator::LaneletPose & spawn_lanelet_pose,
```
hakuturu583, line 51: Please use traffic_simulator::CanonicalizedLaneletPose, functions in API class should use `traffic_simulator::CanonicalizedLaneletPose`

```suggestion
    const traffic_simulator::CanonicalizedLaneletPose & spawn_lanelet_pose,
```

mock/cpp_mock_scenarios/src/cpp_scenario_node.cpp
```diff
@@ -86,7 +95,7 @@ void CppScenarioNode::stop(Result result, const std::string & description)
 }
 
 void CppScenarioNode::spawnEgoEntity(
-  const traffic_simulator::CanonicalizedLaneletPose & spawn_lanelet_pose,
+  const traffic_simulator::LaneletPose & spawn_lanelet_pose,
```
hakuturu583, line 98: Please use traffic_simulator::CanonicalizedLaneletPose, functions in API class should use traffic_simulator::CanonicalizedLaneletPose

```suggestion
  const traffic_simulator::CanonicalizedLaneletPose & spawn_lanelet_pose,
```

mock/cpp_mock_scenarios/include/cpp_mock_scenarios/cpp_scenario_node.hpp
hakuturu583: Propagating a CanonicalizedLaneletPose prevents the possible input of an incorrect LaneletPose.
This change is unacceptable because it works to loosen that check.

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
```diff
@@ -157,13 +165,6 @@ class API
...
-  geometry_msgs::msg::Pose toMapPose(const geometry_msgs::msg::Pose & pose) { return pose; }
-
-  geometry_msgs::msg::Pose toMapPose(const traffic_simulator_msgs::msg::LaneletPose & pose)
-  {
-    return entity_manager_ptr_->getHdmapUtils()->toMapPose(pose).pose;
-  }
-
   template <typename Pose>
   auto spawn(
     const std::string & name, const Pose & pose,
```
hakuturu583, line 170: Member functions of the API class should not accept LanaletPose to prevent incorrect values from being entered by the user, but instead accept CanonicalizedLaneletPose.

hakuturu583 (changes requested):
Member functions of the API class should not accept LanaletPose to prevent incorrect values from being entered by the user, but instead accept CanonicalizedLaneletPose.

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
dmoszynski: @hakuturu583 
Please take a look at that, for the _spawn_ operation, canonicalization takes place in the `EntityManager`, more precisely here:

https://github.com/tier4/scenario_simulator_v2/blob/90b486d9b38739d1b9a1abf2d150eb5df081fc67/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L473-L488

It is worth to note that for canonicalization (`toCanonicalizedLaneletPose`), sometimes the bounding box and matching distance are used. These are not available for an `Entity` that is not spawned yet (it is not in the `EntityManager`). The only thing that can be done in such a situation is to perform a simple canonicalization as in this line.

https://github.com/tier4/scenario_simulator_v2/blob/90b486d9b38739d1b9a1abf2d150eb5df081fc67/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L482

So I believe that the safeguard you mention is already implemented, but the responsibility for this is on the side of the `EntityManager`.

It was developed in such a way that the user of the `API` did not have to keep in mind to perform canonicalization, what's more, to ensure that it is always performed in the same way - no matter if the api is used by `SimulatorCore` or by `CppMockScenarioz`. I believe that, during `spawn()` in `CppMockScenarios` - which is called quite often, it is convenient to simply specify `LaneletPose`, and the responsibility for proper conversion or validation should be strictly inside `traffic_simulator` (since `CanonicalizedLaneletPose` is an "internal type" rather than created to be used for communication - especially due to various canonicalization methods - with bounding box, with different matching distance etc.).

To summarize:
I believe that `spawnEntity()` should use `msg::Pose` and `msg::LaneletPose`. `CanonicalizedLaneletPose` is currently allowed due to the "legacy" side of `SimulatorCore` - it still uses `CanonicalizedLaneletPose`. I believe that the usage of `CanonicalizedLaneletPose` type in `openscenario_interpreter` and `SimulatorCore` can be an unclean and troublesome approach. I believe that we should take care that the conversion between `msg::LaneletPose` and `CanonicalziedLaneletPose` is done in strictly fixed parts and uses as uniform a method as possible (e.g. always the same `matching_distance`). 

Please take a look at [this PR](https://github.com/tier4/scenario_simulator_v2/pull/1252/files), where I plan to introduce more changes related `toCanonicalizeLaneletPose()`.


For this reasons, I believe that passing `msg::LaneletPose` in `API::spawn()` and ensuring that proper canonicalization is performed inside `EntityManager` can help in the future take it in a better direction.

mock/cpp_mock_scenarios/include/cpp_mock_scenarios/cpp_scenario_node.hpp
dmoszynski: @hakuturu583 
Please take a look at [this comment ](https://github.com/tier4/scenario_simulator_v2/pull/1266#discussion_r1643476928)- now the conversion from `msg::LaneletPose` to `CanonicalziedLaneletPose` is the responsibility of `EntityManager::spawnEntity()` - which has access to all the needed data, so there is no need to require `pose::toCanonicalziedLaneletPose()`/`pose::canonicalize()` call before calling `spawn()`.

mock/cpp_mock_scenarios/src/cpp_scenario_node.cpp
dmoszynski: @hakuturu583 
Please take a look at [this comment ](https://github.com/tier4/scenario_simulator_v2/pull/1266#discussion_r1643476928)- now the conversion from `msg::LaneletPose` to `CanonicalziedLaneletPose` is the responsibility of `EntityManager::spawnEntity()` - which has access to all the needed data, so there is no need to require `pose::toCanonicalziedLaneletPose()`/`pose::canonicalize()` call before calling `spawn()`.

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
hakuturu583: > CanonicalizedLaneletPose is currently allowed due to the "legacy" side of SimulatorCore - it still uses CanonicalizedLaneletPose. I believe that the usage of CanonicalizedLaneletPose type in openscenario_interpreter and SimulatorCore can be an unclean and troublesome approach.

That is not true.
CanonicalizedLanaletPose is designed to prevent incorrect values from being entered using the C++ type system and should be used inside scenario_simulator_v2 except when pub/sub is used.

hakuturu583: It also confuses users because functions that use CanonicalizedLanaletPose and traffic_simulator_msgs::msg::LanaletPose inside the API class are in disorder.

In order to prevent incorrect values from being entered and to ensure consistency of API functions, the API class argument type should be CanonicalizedLanaletPose instead of traffic_simulator_msgs::msg::LanaletPose.

dmoszynski: @hakuturu583 
I understand that `CanonicalizedLaneletPose` was introduced to prevent entering incorrect values in `msg::LaneletPose`. I just want to point out that in ss2 code there was a widespread problem with using different arguments  when `CanonicalizedLaneletPose` is created (generally when `msg::Pose` is used).

But you are absolutely right that in the case of conversion `msg::LaneletPose`<->`CanonicalizedLaneletPose` the results will be the same, because in that case `pose::canonicalize()` is used for the conversion, or more precisely simply the [`CanonicalizedLaneletPose()`](https://github.com/tier4/scenario_simulator_v2/blob/ref/RJD-1053-set-update-canonicalized-entity-status/simulation/traffic_simulator/src/data_type/lanelet_pose.cpp) constructor is called, where the `HdMapUtils::canonicalizeLaneletPose` is called:

https://github.com/tier4/scenario_simulator_v2/blob/bb2fe152b2b2be7d926d88dcdb611a460c7cd456/simulation/traffic_simulator/src/hdmap_utils/hdmap_utils.cpp#L157-L179

So conversion `msg::LaneletPose`<->`CanonicalizedLaneletPose` can be successfully done even in `SimulatorCore` or `CppMockScenarios`, but I don't really understand the advantage of "extracting" this conversion outside `traffic_simulator`. 

Could you please let me know if there is any reason why `CanonicalizedLaneletPose` is used for `SimulatorCore`<->`API` communication instead of `msg::LaneletPose`? :bow:  This conversion requires access to lanelet2 (`HdMapUtils`). This introduces a greater dependency between `openscenario_interpreter` and `traffic_simulator`, which can probably be fully avoided by using `msg::LaneletPose` while keeping the code clear. 

Unfortunately, all my knowledge is based on code analysis, and there are certainly arrangements (including those for the future) that I'm unfamiliar with.

Of course, I understand situations such as calling [`entity->getCanonicalizedLaneletPose()`](https://github.com/tier4/scenario_simulator_v2/blob/bb2fe152b2b2be7d926d88dcdb611a460c7cd456/openscenario/openscenario_interpreter/include/openscenario_interpreter/simulator_core.hpp#L151) in `SimulatorCore` and its use, but passing `CanonicalizedLaneletPose` from `openscenario_interpreter` to `traffic_simulator` (e.g., to [set a request](https://github.com/tier4/scenario_simulator_v2/blob/bb2fe152b2b2be7d926d88dcdb611a460c7cd456/simulation/traffic_simulator/src/entity/vehicle_entity.cpp#L183-L203) or spawn) is unclear to me, since `pose::canonicalize()` can be done in `traffic_simulator`, which will help achieve fully my intentions I'm trying to outline below.

From my point of view, it would be good for canonicalization (both `pose::canonicalization()` and `pose::toCanonicalizedLaneletPose()`) to be performed possibly in the same parts of the code. I would also like to add, that with this PR and the [next one](https://github.com/tier4/scenario_simulator_v2/pull/1252/files) - I want to reduce confusion with the conversion of `msg::Pose`/`msg::LaneletPose` to `CanonicalizedLaneletPose`.

I would also like to highlight the issue with multiplicity of canonicalizations. The issue is significant due to the widespread use of `EntityStatus`, where `msg::LaneletPose` occurs, and every time `EntityStatus` is passed and used, another `pose::canonicalize()` must be performed. Of course, the use of `EntityStatus` is unavoidable, but I believe we can make sure that canonicalization is performed a much smaller number of times, in fixed parts of the code and with greater clarity - possibly in the same parts of the code.

For these reasons I've attempted (possibly incorrectly) to unify this and ensure that the `msg::Pose`/`msg::LaneletPose`<->`CanonicalizedLaneletPose` conversion is performed using the attributes of the specific `Entity` - that is, it is performed "close" to the `Entity` - as here:

https://github.com/tier4/scenario_simulator_v2/blob/b9a8c6eb397236a3e06ed9834b8b747f2579935e/simulation/traffic_simulator/src/api/api.cpp#L121-L142

In the [next PR](https://github.com/tier4/scenario_simulator_v2/pull/1252/files), I would like to move `toCanonicalizedLaneletPose()` even more "close" to the Entity (inside `CanonicalizedEntityStatus::set()`:

https://github.com/tier4/scenario_simulator_v2/blob/bb2fe152b2b2be7d926d88dcdb611a460c7cd456/simulation/traffic_simulator/src/entity/entity_base.cpp#L520-L523

https://github.com/tier4/scenario_simulator_v2/blob/bb2fe152b2b2be7d926d88dcdb611a460c7cd456/simulation/traffic_simulator/src/data_type/entity_status.cpp#L66-L84

hakuturu583: > Could you please let me know if there is any reason why CanonicalizedLaneletPose is used for SimulatorCore<->API communication instead of msg::LaneletPose? 🙇 This conversion requires access to lanelet2 (HdMapUtils). This introduces a greater dependency between openscenario_interpreter and traffic_simulator, which can probably be fully avoided by using msg::LaneletPose while keeping the code clear.

I believe that API arguments need to be consistent.
The API should take either LaneletPose or CanonicalizedLaneletPose type arguments, and it is not a good design to accept both for each function.

In the current implementation, only the API::spawn function accepts CanonicalizedLaneletPose, which is inconsistent with other functions.
However, changing all CanonicalizedLaneletPose in API class arguments to LaneletPose would involve a major change, so a separate Pull Request is necessary.
Therefore, in this PR, please revert back to this API::spawn function argument type change.

dmoszynski: @hakuturu583 
Certainly, [done](https://github.com/tier4/scenario_simulator_v2/pull/1266/commits/6d42b00554c71f3c31d8acfaf02b068c446fee5f) ([cpp_mock_scenarios adjustments](https://github.com/tier4/scenario_simulator_v2/pull/1266/commits/6a74a495a028fbb9f711efc91080efa45d98d8bb)). :bow: 

Please let me know if you see needs for any more changes. 
And referring to [this comment](https://github.com/tier4/scenario_simulator_v2/pull/1266#issuecomment-2167731500), please let me know if you recommend rerunning the evaluator.

hakuturu583 (changes requested):
Great work!
But this pull request conflicts to master branch.
Please resolve conflicts.

hakuturu583 (changes requested):
Code looks good to me.
Please re-run regression test in Autoware evaluator.

hakuturu583 (approved):
Thanks!
I confirmed that no regression on this pull request.

dmoszynski:
@yamacir-kit 
Please take a look at this PR again :bow:

yamacir-kit (approved):

PR #1269, opened 2024-05-29
Feature/unit tests/longitudinal speed planner

robomic:
# Description

## Abstract

This PR contains unit tests for `LongitudinalSpeedPlanner`, which is a part of traffic_simulator package.

## Details
All values used for comparisons were calculated by hand.
Function 'getRunningDistance' **is only tested against an estimation**.
All tests are included in `test_longitudinal_speed_planner.cpp` file.
List of tests included in this PR:

1. isAccelerating_true
2. isAccelerating_false
3. isDecelerating_true
4. isDecelerating_false
5. getAccelerationDuration_acceleration
6. getDynamicStates_targetSpeedOverLimit
7. getDynamicStates_maxJerk
8. getDynamicStates_shortAccel
9. isTargetSpeedReached_different
10. isTargetSpeedReached_same
11. getRunningDistance_shortTime
12. getRunningDistance_longTime
13. getRunningDistance_zero

## References
Jira ticket: [internal link](https://tier4.atlassian.net/browse/RJD-1048)

# Destructive Changes

There are no destructive changes.


simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+  auto planner = makeLongitudinalSpeedPlanner();
+
+  geometry_msgs::msg::Twist current_twist{};
+  geometry_msgs::msg::Accel current_accel{};
+  traffic_simulator_msgs::msg::DynamicConstraints constraints{};
+  current_twist.linear.x = 1.0;
+  current_accel.linear.x = 1.0;
+  constraints.max_speed = 10.0;
+  constraints.max_acceleration = 2.0;
+  constraints.max_acceleration_rate = 1.0;
```
hakuturu583, line 103: Please consider using `geometry_msgs::msg::build<message_type>` function for building message.
Avoid creating unnecessary temporary variables for readability.

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+  geometry_msgs::msg::Accel current_accel{};
+  traffic_simulator_msgs::msg::DynamicConstraints constraints{};
+  current_accel.linear.x = 0.0;
+  constraints.max_speed = 20.0;
+  constraints.max_acceleration = 5.0;
+  constraints.max_deceleration = 5.0;
+  constraints.max_acceleration_rate = 5.0;
+  constraints.max_deceleration_rate = 5.0;
+
+  const double epsilon = 1e-8;
```
hakuturu583, line 184: Why do not use https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon ?
And also, please consider using [constexpr](https://en.cppreference.com/w/cpp/language/constexpr)

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+TEST(LongitudinalSpeedPlanner, isTargetSpeedReached_different)
+{
+  auto planner = makeLongitudinalSpeedPlanner();
+
+  geometry_msgs::msg::Twist current_twist{};
+  current_twist.linear.x = 13.0;
+  const double target_speed = 15.0;
+  const double tolerance = 0.1;
```
hakuturu583, line 210: Why do you define temporary variables which use only once.

hakuturu583 (changes requested):

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+{
+  testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
+
+auto makeLongitudinalSpeedPlanner()
+  -> traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner
+{
+  const double step_time = 0.001;
+  const std::string entity = "entity";
```
hakuturu583, line 44: Why do you define temporal variable which was used only once?

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+TEST(LongitudinalSpeedPlanner, isAccelerating_true)
+{
+  auto planner = makeLongitudinalSpeedPlanner();
+
+  geometry_msgs::msg::Twist current_twist{};
+  current_twist.linear.x = 1.0;
+
+  const double target_speed = 2.0;
```
hakuturu583, line 56: Why do you define temporal variable which was used only once?

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+TEST(LongitudinalSpeedPlanner, isAccelerating_false)
+{
+  auto planner = makeLongitudinalSpeedPlanner();
+
+  geometry_msgs::msg::Twist current_twist{};
+  current_twist.linear.x = 3.0;
+
+  const double target_speed = 2.0;
```
hakuturu583, line 67: Why do you define temporal variable which was used only once?

hakuturu583 (changes requested):

hakuturu583:
I will stop the review here because there were a number of points that had similar problems.
Please correct them, including those not specified.

robomic:
Regarding usage of variables that are only used once:
In my opinion explicitly naming these variables is more readable, as we can easily see what is its purpose in an invoked function. To my knowledge, compilers optimize them out.

Please let me know if you prefer:
`  return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{0.001," entity"};`
over this;
`  const double step_time = 0.001;`
`  const std::string entity_name = "entity";`
`  return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{step_time, entity_name};`

If you do prefer, then I will replace all similar examples here and in my other pull requests as well.

robomic:
In [here](https://github.com/tier4/scenario_simulator_v2/pull/1269/commits/70851611c75291fb894b0e44a8da23bacb41f250) I have updated the test description as well as `speed_difference` to be positive.

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,354 @@
...
+  constraints.max_acceleration_rate = 2.0;
+  const double result1_jerk =
+    std::get<2>(planner.getDynamicStates(target_speed, constraints, current_twist, current_accel));
+  EXPECT_DOUBLE_EQ(result1_jerk, constraints.max_acceleration_rate);
+}
+
+/**
+ * @note Test functionality aggregation used in other classes; test calculations correctness
+ * with target_speed slightly smaller than current speed - goal is to test the situation
+ * when the max jerk is not reached because the target_speed is reached first.
```
hakuturu583, line 204: Please change line. Do not split sentence by `;`

```suggestion
 * @note Test functionality aggregation used in other classes
 * Test calculations correctness with target_speed slightly smaller than current speed - goal is to test the situation when the max jerk is not reached because the target_speed is reached first.
```

hakuturu583: Please correct the other sections in the same manner.

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,354 @@
...
+ */
+TEST(LongitudinalSpeedPlanner, isTargetSpeedReached_different)
+{
+  const auto planner = makeLongitudinalSpeedPlanner();
+
+  const auto current_twist = makeTwistWithLinearX(13.0);
+  const double target_speed = 15.0;
+  const double tolerance = 0.1;
+
+  EXPECT_FALSE(planner.isTargetSpeedReached(target_speed, current_twist, tolerance));
```
hakuturu583, line 249:
```suggestion
  EXPECT_FALSE(planner.isTargetSpeedReached(15.0, makeTwistWithLinearX(13.0), 0.1));
```

hakuturu583: Please correct the other sections in the same manner.

hakuturu583 (changes requested):

hakuturu583:
> return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{0.001," entity"};

hakuturu583:
> Regarding usage of variables that are only used once: In my opinion explicitly naming these variables is more readable, as we can easily see what is its purpose in an invoked function. To my knowledge, compilers optimize them out.
> 
> Please let me know if you prefer: ` return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{0.001," entity"};` over this; ` const double step_time = 0.001;` ` const std::string entity_name = "entity";` ` return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{step_time, entity_name};`
> 
> If you do prefer, then I will replace all similar examples here and in my other pull requests as well.

I think this one is better.

```c++
return traffic_simulator::longitudinal_speed_planning::LongitudinalSpeedPlanner{0.001," entity"};
```

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
hakuturu583: > std::numeric_limits::epsilon() is too small and would not make a difference in the machine representation of that number.

I think it is not a machine epsilon, it is a tolerance.
Also, please provide a brief comment explaining why the value was used to determine the success or failure of the test.

robomic:
@hakuturu583
I believe I addressed all the requests.

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  const auto current_accel = makeAccelWithLinearX(0.0);
+  const auto constraints =
+    traffic_simulator_msgs::build<traffic_simulator_msgs::msg::DynamicConstraints>()
+      .max_acceleration(5.0)
+      .max_acceleration_rate(5.0)
+      .max_deceleration(5.0)
+      .max_deceleration_rate(5.0)
+      .max_speed(20.0);
+
+  const double speed_difference = 1e-8;
```
hakuturu583, line 206:
```suggestion
  constexpr double speed_difference = 1e-8;
```

simulation/traffic_simulator/test/src/behavior/test_longitudinal_speed_planner.cpp
```diff
@@ -0,0 +1,324 @@
...
+    .linear(geometry_msgs::build<geometry_msgs::msg::Vector3>().x(x).y(0.0).z(0.0))
+    .angular(geometry_msgs::build<geometry_msgs::msg::Vector3>().x(0.0).y(0.0).z(0.0));
+}
+
+/**
+ * @note Test basic functionality. Test with an accelerating situation.
+ */
+TEST(LongitudinalSpeedPlanner, isAccelerating_true)
+{
+  const auto planner = makeLongitudinalSpeedPlanner();
```
hakuturu583, line 68: Almost all tests create an instance of LongitudinalSpeedPlanner, but this item can be reduced in description by using Gtest's [TestFixture](https://google.github.io/googletest/primer.html#same-data-multiple-tests). 
Please use it to reduce the amount of test code.

hakuturu583 (changes requested):

robomic:
@hakuturu583
I have updated the code.

hakuturu583 (approved):

PR #1275, opened 2024-06-06
Feature: improve ROS parameter handling

TauTheLepton:
# Description

## Abstract

This change improves the ROS parameter obtaining in traffic simulator. It modifies the [`getParameter`][get_parameter] function.
The new function uses `NodeParametersInterface` extracted from a `Node` to quickly access parameters.

## Background

The need for this feature originated from [implementing Traffic Source feature](https://github.com/tier4/scenario_simulator_v2/pull/1231) where it has been noticed that Entity takes over 100ms to spawn when `geometry_msgs::msg::Pose` is used for pose parameter, which was conflicting with high spawn rates of Traffic Source.

## Details

The `getParameter` function is a simple wrapper over the ROS Node parameter interface which can be passed around to different modules. The aim of this change was to eliminate the usage of old [`getParameter`][get_parameter] from traffic simulator. 

The new function is much faster as the `Node` object does not have to be created, only the `Node` object is used to extract the node parameters interface which is later only reused. 
Additionally, the forwarding `getROS2Parameter` function has been added to the API.
The performance has been measured and after the change, spawning an Entity takes a little over 1ms.

### Performance test
There has been a performance test conducted where a total of 200 NPCs were spawned using a modified cpp mock scenario.
100 of them were spawned using `CanonicalizedLaneletPose` and the other 100 were spawned using `geometry_msgs::msg::Pose`.
In both tests, NPCs were spawned in the same location, just using different pose type.

These tests were conducted on both `master` branch and on the branch of this PR.
The results are presented in the tables below.

For both baseline (`master` branch) and evaluation (this PR branch) the log file has been attached showing all measurements in detail.
Please just bear in mind, that _Spawn duration average_ in each individual measurement is an accumulated average of all previous runs and the split average between two pose types has been included at the very end of the file.

#### `master` branch:
|  | `CanonicalizedLaneletPose` | `geometry_msgs::msg::Pose` |
| --- | --- | --- |
| Average spawn time [μs] | 1121.56 | 103277 |
| Max spawn time [μs] | 1720 | 106554 |

[baseline_report.log](https://github.com/user-attachments/files/15891777/baseline_report.log)

#### Branch of this PR
|  | `CanonicalizedLaneletPose` | `geometry_msgs::msg::Pose` |
| --- | --- | --- |
| Average spawn time [μs] | 1215.49 | 1301.36 |
| Max spawn time [μs] | 2098 | 2327 |

[evaluation_report.log](https://github.com/user-attachments/files/15891780/evaluation_report.log)

**Please note that the results above are presented in microseconds.** These times are so low, some deviation should be expected.

## References

[INTERNAL LINK](https://tier4.atlassian.net/browse/RJD-1008)

https://github.com/tier4/sim_evaluation_tools/issues/298

# Destructive Changes

N/A

# Known Limitations

N/A

[get_parameter]: https://github.com/tier4/scenario_simulator_v2/blob/45d42a79d92c370387749ad16c10665deb42e02c/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L56-L65


hakuturu583:
Please complete the regression test before releasing the draft pull request.

simulation/traffic_simulator/include/traffic_simulator/utils/node_parameter_handler.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef TRAFFIC_SIMULATOR__UTILS__NODE_PARAMETER_HANDLER_HPP_
+#define TRAFFIC_SIMULATOR__UTILS__NODE_PARAMETER_HANDLER_HPP_
+
+#include <rclcpp/rclcpp.hpp>
+#include <scenario_simulator_exception/exception.hpp>
+
+namespace traffic_simulator
+{
+class NodeParameterHandler
```
hakuturu583, line 23: Is this class necessary?
I feel that the NodeParameterHandler::getParameter function should also be implemented as a free function that takes node_parameters_interface_ as an argument.

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/api/api.hpp
```diff
@@ -386,6 +388,12 @@ class API
 public:
 #undef FORWARD_TO_ENTITY_MANAGER
 
+  template <typename... Ts>
+  auto getParameter(const std::string & name, Ts &&... xs) const -> decltype(auto)
+  {
+    return node_parameter_handler_->getParameter(name, std::forward<Ts>(xs)...);
+  }
```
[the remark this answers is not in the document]
TauTheLepton: I have changed the name [here](https://github.com/tier4/scenario_simulator_v2/pull/1275/commits/66138a1d749c4d97c177daf975b5afdee138e4d4).

simulation/traffic_simulator/include/traffic_simulator/utils/node_parameter_handler.hpp
TauTheLepton: This class is not strictly necessary. I have implemented the functionality as a class, because when this simple class was passed around it was a nice wrapper over `NodeParametersInterface::SharedPtr` that provided an exact functionality that was needed. When the function received the class pointer as argument, it was clear what this argument was for. Furthermore, if some additional functionality was to be developed in the future, like for example parameter caching, this class would be the place to do this.

I have made the changes you requested and changed the `getParameter` to be a free function in [this commit](https://github.com/tier4/scenario_simulator_v2/pull/1275/commits/2a0cb5aa63e5a7c6a8a9de5e389d8631dfd3f2aa).
If you think the class solution was better, I can revert this commit.

hakuturu583 (approved):

Issue #1276, opened 2024-06-07
Corrected timing of broadcasting entities frames


hakuturu583:
**Is your feature request related to a problem? Please describe.**

The current entities frame averages and broadcasts the coordinates of all entities in every frame.
This causes the visualization frame to jump when an entity spawns, making it difficult to see the rviz.

[Screencast from 06-07-2024 10:57:57 AM.webm](https://github.com/tier4/scenario_simulator_v2/assets/10348912/074e5e82-b3d8-4287-b486-692145be99d2)

**Describe the solution you'd like**

Computing the entities frame in the first frame and then not updating and broadcasting saves computational resources for broadcasting and prevents rviz from becoming difficult to see.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1284, opened 2024-06-12
Optimize `EntityManager::broadcastEntityTransform` to Execute Only Once

TakanoTaiga:
# Description

## Abstract

This pull request modifies the `EntityManager::broadcastEntityTransform()` function to ensure it is called only once.

## Background

The function `broadcastEntityTransform` was previously being called multiple times, causing unnecessary resource usage. This change ensures the function is executed only once, improving efficiency.

## Details

The function `EntityManager::broadcastEntityTransform()` has been modified by introducing a static boolean variable `isSend` to control its execution. Once the function is called, `isSend` is set to true, preventing subsequent calls.

```cpp
void EntityManager::broadcastEntityTransform() {
  static bool isSend;
  if (isSend) {
    return;
  } else {
    isSend = true;
  }

  using math::geometry::operator/;
  using math::geometry::operator*;
  using math::geometry::operator+=;

```

## References

#1276 

# Destructive Changes

N/A

# Known Limitations

N/A

### DIFF
--- a/simulation/traffic_simulator/src/entity/entity_manager.cpp
+++ b/simulation/traffic_simulator/src/entity/entity_manager.cpp
@@ -40,8 +40,11 @@ namespace entity
 void EntityManager::broadcastEntityTransform()
 {
   static bool isSend;
-  if(isSend){return;}
-  else{isSend = true;}
+  if (isSend) {
+    return;
+  } else {
+    isSend = true;
+  }
 
   using math::geometry::operator/;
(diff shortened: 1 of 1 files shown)


hakuturu583 (approved):

hakuturu583 (changes requested):
Code is looks good to me.
Please waiting for evaluator test.

hakuturu583 (approved):

PR #1293, opened 2024-06-20
Optimize entity frame computation.

TakanoTaiga:
# Description

[hoge.webm](https://github.com/tier4/scenario_simulator_v2/assets/53041471/43b6821d-8ea1-42c1-b1be-7790f03b20b6)

## Abstract

This pull request optimizes the entity frame computation by caching the pose and sending it only once, instead of recalculating and broadcasting the coordinates of all entities in every frame.
This Pull Request eliminates the bug that was present in #1284.

## Background

The current implementation causes the visualization frame to jump when an entity spawns, making it difficult to see in rviz. This is due to the averaging and broadcasting of the coordinates of all entities in every frame.

## Details

- Added static variables `is_send` and `pose` to cache the pose and send it only once.
- Updated `broadcastEntityTransform` to use the cached pose for broadcasting.
- This optimization saves computational resources and prevents rviz visualization issues caused by frame jumps.

## References

- #1276
- #1284

# Destructive Changes

N/A

# Known Limitations

N/A


hakuturu583 (changes requested):
Code looks good to me, waiting for evaluator test.

hakuturu583 (approved):

PR #1305, opened 2024-06-28
Feature/publish empty context

hakuturu583:
# Description

## Abstract

Add publish_empty_context parameter to the openscenario_interpreter_node.

## Background

Currently openscenario_interpreter outputs a topic `simulation/context` to visualize the scenario execution status.

However, this functionality has been found to be very slow when the number of Entities and Actions increases.

## Details

I implemented a function to output an empty string to the `simulation/context` topic instead of using this function.

This function works only when the `publish_empty_context` parameter is True. The default value of `publish_empty_context` parameter is False.

This option can be use like below.

```bash
ros2 launch scenario_test_runner scenario_test_runner.launch.py \
     architecture_type:=awf/universe \
     record:=false \
     scenario:='$(find-pkg-share scenario_test_runner)/scenario/sample.yaml' \
     sensor_model:=sample_sensor_kit \
     vehicle_model:=sample_vehicle \
     publish_empty_context:=True
```

## References

N/A

# Destructive Changes

N/A

# Known Limitations

N/A


openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
```diff
@@ -39,13 +39,15 @@ Interpreter::Interpreter(const rclcpp::NodeOptions & options)
...
-  record(false)
+  record(false),
+  publish_empty_context(false)
 {
   DECLARE_PARAMETER(local_frame_rate);
   DECLARE_PARAMETER(local_real_time_factor);
   DECLARE_PARAMETER(osc_path);
   DECLARE_PARAMETER(output_directory);
   DECLARE_PARAMETER(record);
+  DECLARE_PARAMETER(publish_empty_context);
```
yamacir-kit, line 50: Unless you have a technical reason not to, rearrange this line so that it is in lexicographical order with the other DECLARE_PARAMETERs, and also change the declaration order of class data members to be the same as this DECLARE_PARAMETER.

yamacir-kit (changes requested):

openscenario/openscenario_interpreter/src/openscenario_interpreter.cpp
hakuturu583: Fixed in https://github.com/tier4/scenario_simulator_v2/pull/1305/commits/a5a5f8854f3e9e549c8f0594fa0e8dffcfc8225d

hakuturu583: Fixed in https://github.com/tier4/scenario_simulator_v2/pull/1305/commits/6bf5aebe268b71d0fef3e35cedd878abbf3b8ac0

yamacir-kit (approved):

Issue #1312, opened 2024-07-05
Fix Eigen::Vector3d variable definition


MertClk:
## Description

**Parent Bug Report Issue:**
- https://github.com/tier4/scenario_simulator_v2/issues/1302

**Reference to fix:**
- https://eigen.tuxfamily.org/dox/TopicPitfalls.html
> In short: do not use the auto keywords with Eigen's expressions, unless you are 100% sure about what you are doing. In particular, do not use the auto keyword as a replacement for a Matrix<> type. 

### From

```cpp
const auto relative_position =
```

### To

```cpp
const Eigen::Vector3d relative_position =
```

This should fix it.

## Link to breaking PR

- https://github.com/tier4/scenario_simulator_v2/pull/1226#discussion_r1666832342_

hakuturu583:
@MertClk Can you provide information on the operating environment, such as CPU model number/compiler?

We found that this PR implementation changes the behavior of the existing scenario when we check the behavior in the internal regression test environment of TIER IV.

Perhaps the cause of #1312 is a processing or CPU architecture dependent bug due to the Expression Template and the associated computation delay.

Changing the auto type to Eigen::Vector3d seems to change the behavior in some scenarios, and a detailed investigation of the cause and confirmation of the behavior is required to accept this PR.

All computers used in TIER IV are Intel CPUs, gcc compiler and release builds.

MertClk:
@hakuturu583 Below you can find my system information.

* Ubuntu 22.04.4 LTS 64-bit
* Intel Core i9-13900HX
* gcc version 11.4.0
* Build Type: RelWithDebInfo

hakuturu583:
Thank you very much!
I had thought you were using an AMD CPU.

Thanks for the information.

hakuturu583:
@MertClk What eigen version are you using?
Did you install eigen via apt or build eigen by yourself?

MertClk:
@hakuturu583 thank you so much for your support on this issue! I'm using eigen version 3.4.0 and installed via apt.

hakuturu583:
I use same version... :thinking: 
I am currently looking for the cause of the problem considering multiple possibilities, but we are unable to reproduce the problem on our end and fix it quickly.
I apologize for this.

hakuturu583:
It is unclear if it is the model number of the CPU or what is causing the problem,

```bash
const auto relative_position =
````

and

```bash
const Eigen::Vector3d relative_position =
```

[...]

PR #1313, opened 2024-07-05
fix(simple_sensor_simulator): fix eigen variable definition

MertClk:
# Description
Fixes: https://github.com/tier4/scenario_simulator_v2/issues/1302, https://github.com/tier4/scenario_simulator_v2/issues/1312

## Abstract
Fix the definition of eigen variable for calculation of ego vehicle pose.

## Background

In https://github.com/tier4/scenario_simulator_v2/issues/1302, it is reported that kinematic state is not updated.

## References

**Reference to fix:**
- https://eigen.tuxfamily.org/dox/TopicPitfalls.html

This behavior is presumably due to the fact that the Eigen calculation is delayed by the Expression Template, resulting in undefined values. However, a tier4 internal verification showed that this part should not have been a problem even if auto had been used.
One possibility is that the behavior may be CPU architecture dependent.
The cause is unknown, but since there is no particular disadvantage, this PR is merged.

https://github.com/tier4/sim_evaluation_tools/issues/319


hakuturu583:
I am now posting regression test job.

hakuturu583:
I found that release action was failed because of my fault.
So, I ignore failure of release action.

hakuturu583:
Now start running regression test. (Sorry, this is [internal link.](https://github.com/tier4/sim_evaluation_tools/issues/319))

hakuturu583:
Sorry, we are running regression test but it was failed because of something wrong happened in our regression test system.
Now we are try to recovering...
Please wait.

hakuturu583:
I feel very sorry to you.

I have discovered a bug where pull requests from forks cannot be merged due to the Release Action malfunctioning.
Since #1313 is very important and needs to be incorporated urgently, @hakuturu583 will create a similar pull request (#1314) and carry out verification and merging.

hakuturu583:
Fixing release action is working on #1248

PR #1316, opened 2024-07-08
Support `RelativeClearanceCondition` in OpenSCENARIO XML 1.3

HansRobo:
# Description

## Abstract

This pull-request supports `RelativeClearanceCondition` in OpenSCENARIO XML 1.3.
Please refer to the reference for detailed specifications.

image

## Background

In the conventional implementation of `RelativeDistanceCondition`/`DistanceCondition` in scenario_simulator_v2, we were able to distinguish between front and back or left and right by handling negative distance values.
However, these did not conform to the OpenSCENARIO standard.
In order to maintain the expressiveness and maintainability of scenarios while complying with the standard, we decided to implement this RelativeClearanceCondition, which allows you to specify ranges forward, backward, left, and right.

## Details

To implement `RelativeClearanceCondition`, some functions in `traffic_simulator` are implemented.

### `countLaneChanges` function

Previously, you can get lateral distance in lane coodinate.
But its unit is meter and `RelativeClearanceCondition` requires lateral distance in unit "lane".
To address this, a function was implemented which routes between two given points and counts lane changes.
The routing method used in this function is exactly the same as that used in other functions, and internally it corresponds to the implementation for `RoutingAlgorithm::undefined` / `RoutingAlgorithm::shortest`.
`RelativeClearanceConditon` implemented using the function implementation for `RoutingAlgorithm::shortest`.

## References

- [Class RelativeClearanceCondition | ASAM OpenSCENARIO v1.3.0](https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_XML/latest/generated/content/RelativeClearanceCondition.html)
- [Regression Test: OK(latest: 8/20)](https://github.com/tier4/sim_evaluation_tools/issues/322)

# Destructive Changes

None

# Known Limitations

The `oppositeLanes` option is not supported in this pull-request.
This is related to the fact that the current implementation of traffic_simulator does not generally support routing that includes opposite lanes.


openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/integer.hpp
```diff
@@ -35,6 +35,8 @@ struct Integer
 
   explicit Integer(const std::string &);
 
+  static auto infinity() noexcept -> Integer;
```
yamacir-kit, line 38: As std::numeric_limits does not define a member function infinity for integer types, it is not possible to represent infinity with integer types. Please modify it to provide min and max.

openscenario/openscenario_interpreter/src/syntax/relative_clearance_condition.cpp
```diff
@@ -0,0 +1,147 @@
...
+            triggering_entities.entity_refs.begin(), triggering_entities.entity_refs.end(),
+            [&](const auto & triggering_entity) {
+              return triggering_entity.name() == entity.first;
+            }) == triggering_entities.entity_refs.end()) {
+          entities.emplace_back(entity.first);
+        }
+      }
+    }
+    return entities;
+  }()),
```
yamacir-kit, line 49: Why is the initialization of entity_refs implemented like this? It seems that the behavior is programmed differently from the conventions in the interpreter and from the OpenSCENARIO XML standard.

Also, since EntitySelection is supported, the name referenced by EntityRef may be an alias that points to multiple entities, so each syntax should be stored in Entity type instead of EntityRef. For more information, please see the following pull request: https://github.com/tier4/scenario_simulator_v2/pull/1059

yamacir-kit (changes requested):

openscenario/openscenario_interpreter/src/syntax/relative_clearance_condition.cpp
HansRobo: > Why is the initialization of entity_refs implemented like this? It seems that the behavior is programmed differently from the conventions in the interpreter and from the OpenSCENARIO XML standard.

What is exactly "programmed differently from the conventions in the interpreter and from the OpenSCENARIO XML standard"?
I didn't get it, so there may be problems with my understanding of the conventions and starndards.
Could you tell me your expectations and the difference of current code?

Note: Lines from 36 to 47 are implemented for the standard description `If it is not used then all entities are considered.` for `entityRef`.
Here, this specification is performed at the same time as initialization and written to the `entity_refs` member.
Is what is being pointed out here that the problem is that what is actually written in the scenario and the result of class construction are different?

> Also, since EntitySelection is supported, the name referenced by EntityRef may be an alias that points to multiple entities, so each syntax should be stored in Entity type instead of EntityRef. For more information, please see the following pull request: https://github.com/tier4/scenario_simulator_v2/pull/1059

OK. I'll fix.

yamacir-kit: > Note: Lines from 36 to 47 are implemented for the standard description If it is not used then all entities are considered. for entityRef.

If you want to consider all entities, you should select entities that are `is_added` out of global()->entities. The code you wrote is "global()->entities whose names are contained in triggering_entities.entitiy_refs". So my question is what is the need to touch triggering_entities.entity_refs to search for "all entities".

HansRobo: > you should select entities that are is_added out of global()->entities

Sorry, I didn't know this was necessary to filter with `is_added`. I'll try adding a filter by `is_added`.

> The code you wrote is "global()->entities whose names are contained in triggering_entities.entitiy_refs".

This is the opposite.
Note that the operator used in the comparison between `find_if` and `end` is "==", not "!=".
(The fact that you made a mistake here might indicate that my code is difficult to read.)

> Also, the number of "all entities" varies since the construct of this class, considering the existence of AddEntityAction and DeleteEntityAction. Therefore, I think that the search for "all entities" should be performed each time a condition is determined.

Oh, I missed it. Thank you for pointing out my missed cases!

yamacir-kit: > This is the opposite.
Note that the operator used in the comparison between find_if and end is "==", not "!=".
(The fact that you made a mistake here might indicate that my code is difficult to read.)

I was wrong about this, sorry.

openscenario/openscenario_interpreter/include/openscenario_interpreter/syntax/integer.hpp
HansRobo: fixed in [`29bbac4` (#1316)](https://github.com/tier4/scenario_simulator_v2/pull/1316/commits/29bbac4258f0dd4c4c5381d959bd354ccc46d0dc)

yamacir-kit (approved):

PR #1318, opened 2024-07-10
add Longitudinal_control.md

curious-jp:
# Description
## Abstract

Added document for longitudinal control of entities. 

## Background

Since there was no documentation organizing about longitudinal control of entities, we have added one.

## Details

Written document about longitudinal control API given to users.

## References

N/A

# Destructive Changes

N/A.  

# Known Limitations

N/A


hakuturu583 (changes requested):
Please also describe how these function works. (e.g The requestSpeedChange function indicates the target speed, but since the speed is the target speed, it should be stated that the upper speed limit specified by the setVelocityLimit function, etc. must be observed.)

hakuturu583 (approved):

PR #1233, opened 2024-04-16
ref(traffic_simulator): implement separate class for distance calculations, adapt make positions in SimulatorCore

dmoszynski:
# Description

## Abstract
This PR is intended to separate methods related to distance calculation from `API`/`EntityManager`/`EntityBase` modules.

Methods that have been completely separated in this PR:
- `getLateralDistance()`
- `getLongitudinalDistance()`
- `getBoundingBoxDistance()`
- `getBoundingBoxLaneLateralDistance()`
- `getBoundingBoxLaneLongitudinalDistance()`
- `getDistanceToLaneBound()`
- `getDistanceToLeftLaneBound()`
- `getDistanceToRightLaneBound()`
- `getDistanceToStopLine()`
- `getDistanceToCrosswalk()`

The result is a collection of functions in namespace [traffic_simulator::distance](https://github.com/tier4/scenario_simulator_v2/pull/1233/files#diff-d1629bc83ad89d40c81f4d6e67fb8c5130684910b5189cd323b203e57ddbf51b).

Methods that have been separated because they use `traffic_simulator::distance` (they are called by `DistanceCondition` etc.) and their separation was already advantageous for "approach validity". Separation of more methods related to "Pose" will be introduced in the next PR ([RJD-1053](https://tier4.atlassian.net/browse/RJD-1053)):
- `getRelativePose()`
- `getBoundingBoxRelativePose()`
- `makeNativeRelativeWorldPosition()`
- `makeNativeBoundingBoxRelativeWorldPosition()`
- `makeNativeRelativeLanePosition()`
- `makeNativeBoundingBoxRelativeLanePosition()` 

The result is a collection of functions in namespace [traffic_simulator::pose](https://github.com/tier4/scenario_simulator_v2/pull/1233/files#diff-95bb7bd23a79b9381b4a8359f6d38fe6af3e87c124c82743135a69b7498ac31f).

Note: I realize that I've hurried to separate this pose-related functions in this PR, if you feel that it makes the PR too big - I will try to split it. However, I took into account here that if I don't do it now then in the next PR I will have to modify the same code basically. Moreover, considering how many files will be changed in PR ([RJD-1053](https://tier4.atlassian.net/browse/RJD-1053)) (separating `toLaneletPose()`, `canonicalize()`, `toMapPose()` etc. - they are very commonly used), making this change now can be useful for reducing the size and complexity of the next PR [RJD-1053](https://tier4.atlassian.net/browse/RJD-1053).

## Background
[...]

PR #1320, opened 2024-07-10
ref(behavior_tree, traffic_simulator): move responsibility for canonicalization to traffic_simulator, simplify 

dmoszynski:
# Description

This PR is based on [TMP_PR2](https://github.com/tier4/scenario_simulator_v2/pull/1252) - for which regression tests have been run and no regression has been [evidenced](https://github.com/tier4/sim_evaluation_tools/issues/281#issuecomment-2118251637).

## Abstract
This PR is intended to remove the `EntityStatus` canonicalization process from `BehaviorTree`. From now on, during an update, for each Entity the canonicalization is performed only once and takes place only in `EntityBase::onUpdate()`.

In addition, many unnecessary `EntityStatus` canonicalizations have been removed in the `traffic_simulator` itself. Now the `CanonicalizedEntityStatus` object is created when the entity is spawned, then the setting of the status is done via `EntityBase::setStatus()` where the EntityStatus is passed, then the `CanonicalizedEntityStatus::set()` method is called - where the desired canonicalization takes place.

Note: This PR is a continuation of the refactoring of the canonicalization process, which was initiated by the [previous PR](https://github.com/tier4/scenario_simulator_v2/pull/1266).

## Background

At present, there are many calls of the `CanonicalizedEntityStatus` constructor in the code. The canonicalization process takes place in many places, which can be simplified - in fact, it is required that the status set in the update result is canonicalized - this is guaranteed by this PR.

## Details

Canonicalization_description drawio(1)

### Canonicalization of EntityStatus in TrafficSimulator
- removed `CanonicalizedEntityStatus` constructors that accept `EntityStatus`, 
- removed overload of `oparator=` in `CanonicalizedEntityStatus`,
- `set()` method added to `CanonicalizedEntityStatus` (it is important to set `lanelet_pose_valid = true/false` in passed `EntityStatus`, this determines whether the canonicalization is performed using `LaneletPose` itself or `msg::Pose` is used.)
- `EntityStatus` canonicalization process moved to `API::setEntityStatus`->`EntityBase::setStatus()`->`CanonicalizedEntityStatus::set()`

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/src/data_type/entity_status.cpp#L66-L84

### EntityBase improvements
- changed the accepted `EntityBase::setStatus()` argument to `EntityStatus` and added `EntityStatus::setCanonicalizedStatus()` with `CanonicalizedEntityStatus` argument for cases where `CanonicalizedLaneletPose` is passed from `openscenario_interpreter`

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/src/entity/entity_base.cpp#L529-L537

- the setters in `EntityBase` have  improved, now setters from `CanonicalizedEntityStatus` are used directly, there is no unnecessary process of taking `EntityStatus` and canonicalizing it again.

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/src/entity/entity_base.cpp#L555-L567

- `getEntityType()` method  added to `EntityBase`
- grouped and ordered methods in `CanonicalizedEntityStatus`

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/include/traffic_simulator/data_type/entity_status.hpp#L40-L71

- added `EgoEntity::setStatus()` with `lanelet::Ids` as second argument, since it is the only Entity that uses `lanelet_id` not from `EntityStatus` but got from `Autoware`

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/src/entity/ego_entity.cpp#L299-L318

- It is ensured that EntityManager::getEntityStatus() simply returns a constant reference to the map object

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/src/entity/entity_manager.cpp#L178-L186

### Canonicalization of EntityStatus in BehaviorTree 
- Changed "updated_status" port type in `BehaviorTree` from `CanonicalizedEntityStatus` to `EntityStatus`, port name now is "non_canonicalized_updated_status".

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/include/traffic_simulator/behavior/behavior_plugin_base.hpp#L75

- Now `ActionNode::calculateUpdatedEntityStatus` and `ActionNode::calculateUpdatedEntityStatusInWorldFrame` return `EntityStatus`
- removed `fill_lanelet_data_and_adjust_orientation` from `FollowTajetoryAction`, now every `CanonicalizedEntityStatus` has a fitted position to the lanelet
- It is ensured that `ActionNode::getEntityStatus()` simply returns a constant reference to the map object

# Destructive Changes

## BehaviorTree
- Changed "updated_status" port type in `BehaviorTree` from `CanonicalizedEntityStatus` to `EntityStatus`, port name now is "non_canonicalized_updated_status". 

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/traffic_simulator/include/traffic_simulator/behavior/behavior_plugin_base.hpp#L75

#### Before

https://github.com/tier4/scenario_simulator_v2/blob/140ada513a9b78fc9bffffb76e142f55ed1c4b75/simulation/behavior_tree_plugin/src/vehicle/follow_lane_sequence/yield_action.cpp#L124-L126

#### After

https://github.com/tier4/scenario_simulator_v2/blob/bbe65a1611b1c50a3c54f0058c102068add5471f/simulation/behavior_tree_plugin/src/vehicle/follow_lane_sequence/yield_action.cpp#L124-L126


hakuturu583:
I think `toCanonicalizedLaneletPose` function should be `pose::toCanonicalizedLaneletPose`.

simulation/traffic_simulator/include/traffic_simulator/data_type/entity_status.hpp
```diff
@@ -17,60 +17,69 @@
...
+
   auto getLinearJerk() const noexcept -> double;
-  auto setTime(double) -> void;
-  auto getTime() const noexcept -> double;
+  auto setLinearJerk(double) -> void;
+
+  auto laneMatchingSucceed() const noexcept -> bool;
+  auto getLaneletId() const noexcept -> lanelet::Id;
+  auto getLaneletIds() const noexcept -> lanelet::Ids;
+  auto getLaneletPose() const noexcept -> LaneletPose;
```
yamacir-kit, line 66:
```suggestion
  auto getLaneletPose() const noexcept -> const LaneletPose &;
```

simulation/traffic_simulator/include/traffic_simulator/data_type/entity_status.hpp
```diff
@@ -17,60 +17,69 @@
...
+  auto getCanonicalizedLaneletPose() const noexcept -> std::optional<CanonicalizedLaneletPose>;
```
yamacir-kit, line 67:
```suggestion
  auto getCanonicalizedLaneletPose() const noexcept -> const std::optional<CanonicalizedLaneletPose> &;
```

mock/cpp_mock_scenarios/src/traffic_source/define_traffic_source_delay.cpp
```diff
@@ -75,19 +75,17 @@ class DefineTrafficSourceDelay : public cpp_mock_scenarios::CppScenarioNode
         stop(cpp_mock_scenarios::Result::FAILURE);  // LCOV_EXCL_LINE
       }
       for (const auto & name : names) {
-        const traffic_simulator::CanonicalizedEntityStatus entity_status =
-          api_.getEntityStatus(name);
+        if (const auto entity = api_.getEntity(name)) {
+          const bool is_vehicle =
+            entity->getEntityType().type == traffic_simulator_msgs::msg::EntityType::VEHICLE;
```
hakuturu583, line 80: The same process is implemented in many scenarios.
For ease of maintenance, please implement it in the [base class for C++ scenarios](https://github.com/tier4/scenario_simulator_v2/blob/master/mock/cpp_mock_scenarios/include/cpp_mock_ scenarios/cpp_scenario_node.hpp) file.

hakuturu583: Please correct many of the same variables for scenarios that we have not pointed out.

simulation/behavior_tree_plugin/include/behavior_tree_plugin/action_node.hpp
```diff
@@ -88,26 +89,26 @@ class ActionNode : public BT::ActionNodeBase
       BT::InputPort<std::shared_ptr<traffic_simulator::TrafficLightManager>>("traffic_light_manager"),
       BT::InputPort<traffic_simulator::behavior::Request>("request"),
       BT::OutputPort<std::optional<traffic_simulator_msgs::msg::Obstacle>>("obstacle"),
-      BT::OutputPort<std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus>>("updated_status"),
+      BT::OutputPort<std::shared_ptr<traffic_simulator::EntityStatus>>("non_canonicalized_updated_status"),
```
hakuturu583, line 92: Why do you passing uncanonicalized entity status?
I read pull-request description `Canonicalization of EntityStatus in BehaviorTree` section, but I could not understand why we have to change port type.

simulation/behavior_tree_plugin/include/behavior_tree_plugin/pedestrian/behavior_tree.hpp
```diff
@@ -68,7 +68,7 @@ class PedestrianBehaviorTree : public BehaviorPluginBase
   DEFINE_GETTER_SETTER(StepTime,                                         double)
   DEFINE_GETTER_SETTER(TargetSpeed,                                      std::optional<double>)
   DEFINE_GETTER_SETTER(TrafficLightManager,                              std::shared_ptr<traffic_simulator::TrafficLightManager>)
-  DEFINE_GETTER_SETTER(UpdatedStatus,                                    std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus>)
+  DEFINE_GETTER_SETTER(UpdatedStatus,                                    std::shared_ptr<traffic_simulator::EntityStatus>)
```
hakuturu583, line 71: Inside traffic_simulator, it is desirable to reduce exception handling by taking canonicalized LaneletPose/EntityStatus as arguments and return values whenever possible.
Therefore, please specify the circumstances under which this change must be resolved.

hakuturu583 (changes requested):
The CanonicalizedEntityStatus/LaneletPose type is designed to exchange normalized lane coordinate system information inside traffic_simulator.
All EntityStatus/LaneletPose types used in the internal processing of traffic_simulator must be of type CanonicalizedEntityStatus/LaneletPose.

yamacir-kit (changes requested):

simulation/traffic_simulator/include/traffic_simulator/data_type/entity_status.hpp
dmoszynski: @yamacir-kit 
Thank you so much, my overlook, I've [provided const reference where it was missing](https://github.com/tier4/scenario_simulator_v2/pull/1320/commits/850f40441c8727d8ca55705c9ff3ea4354edf680#diff-541c9fdcaf8e81f7ab6d1fb2a82521c2de7fb693b30cb8c466cadeedde341757) in `CanonicalizedEntityStatus`. :bow:

simulation/traffic_simulator/include/traffic_simulator/data_type/entity_status.hpp
dmoszynski: @yamacir-kit 
Thank you so much, my overlook, I've [provided const reference where it was missing](https://github.com/tier4/scenario_simulator_v2/pull/1320/commits/850f40441c8727d8ca55705c9ff3ea4354edf680#diff-541c9fdcaf8e81f7ab6d1fb2a82521c2de7fb693b30cb8c466cadeedde341757) in `CanonicalizedEntityStatus`. :bow:

mock/cpp_mock_scenarios/src/traffic_source/define_traffic_source_delay.cpp
dmoszynski: @hakuturu583 
Certainly, [here is commit](https://github.com/tier4/scenario_simulator_v2/pull/1320/commits/8425b2545bfe4cbd0819941fe9df3b8f909ce7e1) introduces changes :bow:

dmoszynski:
@hakuturu583 

> I think `toCanonicalizedLaneletPose` function should be `pose::toCanonicalizedLaneletPose`.

Namespace `pose` is defined as “`inline`” according [to this recommendation](https://github.com/tier4/scenario_simulator_v2/pull/1233#discussion_r1591884371) - so it is allowed to use `toCanonicalizedLaneletPose` instead of `pose::toCanonicalizedLaneletPose`. However, I agree that `pose::` is more clear - that's why I've introduced the changes that [this commit contains](https://github.com/tier4/scenario_simulator_v2/pull/1320/commits/d940ab04ba3a67fd5e5351136ec02236623d0914).

simulation/behavior_tree_plugin/include/behavior_tree_plugin/pedestrian/behavior_tree.hpp
dmoszynski: @hakuturu583 
### Create CanonicalizedEntityStatus
In the approach proposed in this PR, the `CanonicalizedEntityStatus` object is created for a specific entity **only once** - during its spawn, more specifically here: [EntityManager::spawnEntity()](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/include/traffic_simulator/entity/entity_manager.hpp#L473-L486). 
During the spawn, an `EntityStatus` object and a `CanonicalizedLaneletPose` object are created based on the arguments passed (parameters, pose etc.), these two objects forming the `CanonicalizedEntityStatus`. 
Created `CanonicalizedEntityStatus` is passed to the constructor of `EntityBase`. 

### Access to data in CanonicalizedEntityStatus
`EntityBase` has getters providing access to unmodifiable `CanonicalizedEntityStatus` data and setters allowing safe modification of selected objects. 
Especially it is impossible to modify `LaneletPose` itself. The only way to change the `CanonicalizedEntityStatus` (so also `LaneletPose`) is to pass the new `EntityStatus` to [`EntityBase::setStatus()`](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/entity/entity_base.cpp#L532-L541).
This method calls [`CanonicalizedEntityStatus::set()`](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/data_type/entity_status.cpp#L56-L91), where **canonicalization is performed** based on the passed `EntityStatus`. In such an architecture, it is not possible for `EntityBase` to have invalid `LaneletPose` (non-canonicalized) data.
Moreover, if the canonicalization fails, `CanonicalizedLaneletPose` simply will not have a value (it will be `std::nullopt`) - so the exception will not be thrown. 
Therefore, there is no need to handle exceptions in the case of canonicalization, only the `EntityBase::laneMatchingSucceed()` check is required - as it is developed for a long time in many parts of the code. 

### Advantages
The major advantage of this approach is that **during each tick** of the simulator, after the `BehaviorTree` updates the entity state (`EntityStatus`), canonicalization is performed **only once in one place**, to be more precise, for example, here: [VehicleEntity::onUpdate](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/entity/vehicle_entity.cpp#L156).
Thanks to the fact that the canonicalization is performed in **one place**, on the existing `CanonicalizedEntityStatus`, it is always performed in the same way and in case `LaneletPose` is invalid, the objects needed for its setting are [available in advance](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/data_type/entity_status.cpp#L79-L81): 
- `BoundingBox` from `CanonicalizedEntityStatus` is used, 
- `matching_distance` defined in `EntityBase::getDefaultMatchingDistanceForLaneletPoseCalculation()` is used,
- `EntityType` from `CanonicalizedEntityStatus` is used to deduce `include_crosswalk`.

Moreover, performing canonicalization using the previously existing `CanonicalizedEntityStatus`, helps to ensure that during the update some values of the `EntityStatus` [have not changed](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/data_type/entity_status.cpp#L58-L63) (for example, the `BoundingBox` should be constant) - so it is **safe canonicalization**.

Thanks to the fact that canonicalization is performed in one place, we have a clear **division of responsibility** - from now on it is the full responsibility of `EntityBase` and its `EntityBase::setStatus()` methods to ensure that the data in `EntityBase->CanonicalizedEntityStatus->EntityStatus` are always correct (also canonicalized always in the same way).

## Circumstances
To achieve this (**safe canonicalization only once per tick in one place**), it is required to use `EntityStatus` as the result of an entity status update in `BehaviorTree`. Otherwise, canonicalization would be performed in `BehaviorTree` - in many places, in a way that is not always the same - in the implementation before this PR, there was a problem with `matching_distance`, which occurs in `EntityBase` and was not passed to `BehaviorTree`, which resulted in different values during canonicalization. 

Leaving the creation of `CanonicalizedEntityStatus` in `BehaviorTree` makes canonicalization unclear, can be unsafe, requires more resources, and it is unknown which module is in fact responsible for it.

To put it in other words, if `CanonicalizedEntityStatus` is created in other places, e.g. in `BehaviorTree`, the advantages from this PR will not be as great as it could be.

## Graph
Here I've attached a graph showing the architecture that this PR introduces.
Canonicalization_description drawio

####  CanonicalizedLaneletPose  OpenScenarioInterpreter->API
As you can see in the graph there are 2 methods marked in red:
- `auto EntityBase::setCanonicalizedStatus(const CanonicalizedEntityStatus &) -> void;`  
- `auto CanonicalizedEntityStatus::set(const CanonicalizedEntityStatus & status) -> void;`

These are methods that cannot be eliminated at this time, since the communication between `OpenScenarioInterpreter` and the `API` is done via `CanonicalizedLaneletPose`, especially the [API::setEntityStatus()](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/api/api.cpp#L107C1-L121) method accepts this data type - these are methods are used only for this purpose. 
This causes canonicalization to be performed on the `OpenScenarioInterpreter` side, which in my feeling should also be eliminated, and I believe it will not be difficult to achieve. 

To put it in other words, we should aim to make the canonicalization of `EntityStatus` happen in the [`EntityBase::setStatus()`](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/entity/entity_base.cpp#L532-L541) method and only here.

hakuturu583: > EntityBase has getters providing access to unmodifiable CanonicalizedEntityStatus data members and setters allowing safe modification of selected data members.
Especially it is impossible to modify LaneletPose itself. The only way to change the CanonicalizedEntityStatus (so also LaneletPose) is to pass the new EntityStatus to [EntityBase::setStatus()](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/entity/entity_base.cpp#L532-L541).
This method calls [CanonicalizedEntityStatus::set()](https://github.com/tier4/scenario_simulator_v2/blob/01f9907aa600be824d8e6230e7bf0c9da87acf92/simulation/traffic_simulator/src/data_type/entity_status.cpp#L56-L91), where canonicalization is performed based on the passed EntityStatus.
In such an architecture, it is not possible for EntityBase to have invalid LaneletPose (non-canonicalized) data.
Moreover, thanks to the fact that canonicalization takes place in one place, catching exceptions should be easier, especially since canonicalization is always done on the TrafficSimulator side.
It is worth noting that the only exception that can occur is in the case of lanelet_pose_valid == True and the trouble with canonicalization exactly [here](https://github.com/tier4/scenario_simulator_v2/blob/127bbaefb4d101e00828cd6fc6585cf76b9e03be/simulation/traffic_simulator/src/data_type/entity_status.cpp#L76) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/127bbaefb4d101e00828cd6fc6585cf76b9e03be/simulation/traffic_simulator/src/utils/pose.cpp#L55) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/127bbaefb4d101e00828cd6fc6585cf76b9e03be/simulation/traffic_simulator/src/data_type/lanelet_pose.cpp#L81-L88).

This is just the way it is currently implemented.
It is common in programming for assumptions to be unintentionally broken when adding functionality.
It is necessary to continue to ensure that types are canonicalized in order to reduce the risk of introducing bugs in the future.

hakuturu583: I think it is important to have one place to implement Canonicalize, but it would be better if it were just one function to Canonicalize.
If there are too many functions that take as arguments or return values LeneletPose/EntityStatus that have not been Canonicalized that may contain invalid values, additional error handling will be required in various places, which will increase maintenance costs.
Therefore, this implementation should be modified.

dmoszynski: @hakuturu583 

> This is just the way it is currently implemented.
> It is common in programming for assumptions to be unintentionally broken when adding functionality.
> It is necessary to continue to ensure that types are canonicalized in order to reduce the risk of introducing bugs in the future.

I fully agree that before this PR this was also guaranteed - in every place where such a problem could occur additional `CanonicalizedEntityStatus` objects were created - which solves the problem of invalid values, of course, but in my feelings it is not very efficient. I also want to point out the common occurrence in the code before PR of `static_cast<EntityStatus>(status)` casting and re-creating another `std::shared_ptr<CanonicalziedEntityStatus>` objects - which can be avoided without increasing the risk of invalid values - as developed in this PR.

Especially since, the methods from before this PR never operate on `CanonicalizedEntityStatus`, only in the line where the result is returned canonicalization takes place. In almost entire `BehaviorTree`, operations are performed in such a way that canonicalized input data are cast to `EntityStatus`, calculations are performed using `LaneletPose`, and canonicalization takes place only at return - once the object passed to `TrafficSimulator` is created - so at the end of each method performing the action.

https://github.com/tier4/scenario_simulator_v2/blob/39594fd362c91b32514863d4ee02b2d4dcaaf69a/simulation/behavior_tree_plugin/src/vehicle/lane_change_action.cpp#L263-L267

This PR only removes the creation of `CanonicalizedEntityStatus` in the return line of every method, but guarantees that this object is created in `EntityBase` after receiving the result from `BehaviorTree`.

I think that removing the creation of `CanonicalizedEntityStatus` in `BehaviorTree`, or indeed moving it to `EntityBase::setStatus()`, does not increase the chances of introducing bugs, but decreases them - since it guarantees that canonicalization will certainly always be performed before setting the status in `EntityBase`, and certainly in the already developed, uniform and proper way. 

> If there are too many functions that take as arguments or return values LeneletPose/EntityStatus that have not been Canonicalized that may contain invalid values, additional error handling will be required in various places, which will increase maintenance costs.

I believe that an approach in which `EntityStatus` and `LaneletPose` are allowed as input to `TrafficSimulator`, instead of their canonicalized equivalents will in fact protect us from the need to handle exceptions in multiple places. For example, every time `pose::canonicalize()` is called to return a result at the end of the methods in the `ActionNode` code, there is a risk of throwing an exception in those places, leading to the need to handle it.  Moving canonicalization to one place will allow catching errors in one place.

It is possible that I overlooked something and you mean a completely different error handling, if it is so please guide me. :bow: 

#### Summary

I fully agree that operations inside `TrafficSimulator` should be performed using `CanonicalizedLaneletPose` and `CanonicalizedEntityStatus` (as it is). I also agree that the input data for `BehaviorTree` should be canonicalized (as it is). However, I do not see any obstruction in making `TrafficSimulator` itself responsible for canonicalization of input data (also including those from `OpenScenarioInterpreter`). In other words, in my feelings, input data verification should be performed in `TrafficSimulator` itself. 

### Alternative solution

I've just developed an alternative solution and fully tested it - it works as expected. 

It is based on the fact that the communication between  `TrafficSimulator` and  `BehaviorTree` uses a shared_ptr. Therefore, nothing prevents `BehaviorTree` from writing data to the same `CanonicalizedEntityStatus` object that is in   `EntityBase`. To put it in other words - instead of returning the `EntityStatus` from `BehaviorTree` to `TrafficSimulator` it simply writes the result of the action directly to `CanonicaliedEntityStatus` - which is shared between `BehaviorTree` and `TrafficSimulator`. I think this solution is clear and efficient, but I'm not sure about its maintenance safety - I'm quite not a fan of using shared_ptr for such communication. 

Please let me know what you think about it. I've included the changes [in this summary commit](https://github.com/tier4/scenario_simulator_v2/pull/1320/commits/de4c942b8f2ce234b4a6bc95e457e7ed3df89558), it is quite large, but I hope it is readable. :bow: :bow: 

Example: [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/behavior_tree_plugin/src/pedestrian/walk_straight_action.cpp#L50) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/behavior_tree_plugin/src/action_node.cpp#L97-L102) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/traffic_simulator/include/traffic_simulator/behavior/behavior_plugin_base.hpp#L58) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp#L250) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/traffic_simulator/src/entity/pedestrian_entity.cpp#L257) -> [here](https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/traffic_simulator/src/entity/pedestrian_entity.cpp#L260-L261)

hakuturu583: > In other words, in my feelings, input data verification should be performed in TrafficSimulator itself.

I do not think so.
This is because the behavior plugin may plan in the lane coordinate system or in the map coordinate system, depending on the implementation of the behavior.
Instead of calculating and returning the “Next EntityStatus” inside the behavior plugin, it would be better to change the interface to “return the amount of movement” and do the lane coordinate system calculation inside the behavior plugin.

hakuturu583:
```cpp
class Distance {
};

class LaneCoordinateDistance : public Distance {
};

class CartesiaonCoordinateDistance : public Distance {
};

class DeltaStatus
{
    std::unique_ptr<Distance> delta_distance;
    // traffic_simulator_msgs::msg::ActionStatus delta_action;
};

class CanonicalizedEntityStatus
{
    CanonicalizedEntityStatus operator+( DeltaStatus &other );
};

CanonicalizedEntityStatus CanonicalizedEntityStatus::operator+( DeltaStatus &other ){
    // Implement a function to calculate the next CanonicalizedEntityStatus considering DeltaStatus
};
```

I recommend expressing distances and coordinate systems as types as shown below.
If designed this way, there is no need to generate CanonicalizedEntityStatus many times, and the Canonicalize process will not be complicated by exception handling in both traffic_simulator/behavior_plugin.

dmoszynski: @hakuturu583 

> I do not think so.
> This is because the behavior plugin may plan in the lane coordinate system or in the map coordinate system, depending on the implementation of the behavior.

The effect of `BehaviorTree` action can indeed be reflected in the `EntityStatus` in one of the two coordinate systems, but we only consider two cases at the moment:
- the movement is planned in the map coordinate system (msg::Pose variable), the `lanelet_pose_valid` variable is set to **False** -  `lanelet_pose` variable is calculated based on pose with the canonicalized value inside `CanonicalizedEntityStatus::set()`
- movement is planned in the lanelet coordinate system (msg::LaneletPose variable), the `lanelet_pose_valid` variable is set to **True** - `lanelet_pose` variable is currently filled, only canonicalization is performed inside `CanonicalizedEntityStatus::set()` - here it is very essential that the pose variable (msg::Pose) also has to be filled in the `EntityStatus` received by `TrafficSimulator`

It can be seen here:
https://github.com/tier4/scenario_simulator_v2/blob/de4c942b8f2ce234b4a6bc95e457e7ed3df89558/simulation/traffic_simulator/src/data_type/entity_status.cpp#L66-L84

I feel that the above approach (from this PR) represents almost what you wrote using `DeltaStatus`, with one significant difference. Your idea handles the case when `lanelet_pose` is filled (`lanelet_pose_valid` variable is set to **True**), but `pose` (msg::Pose) is not filled - now such a case cannot occur and `pose` must always be filled. 
For this reason, I think your approach is really good and worth implementing, but I'm afraid not within this PR - it would introduce too many changes and possible troubles.

Could you please take a look at the [_Alternative Solution_ ](https://github.com/tier4/scenario_simulator_v2/pull/1320#discussion_r1695775781) presented above and tell me what you think about it?

dmoszynski: @hakuturu583 
Naturally, if the [Alternative Solution](https://github.com/tier4/scenario_simulator_v2/pull/1320#discussion_r1695775781) is not as expected, let me know :bow: :bow: 
 then I will revert the changes and bring back the creation of the `CanonicalizedEntityStatus` on the `BehaviorTree` side and sending it to the `TrafficSimulator` (as it was).

hakuturu583: > I feel that the above approach (from this PR) represents almost what you wrote using DeltaStatus, with one significant difference.

Hmmm, I don't feel that way.
I think the two advantages of my proposal are that the Canonicalize process is no longer needed inside the behavior plugin, and the C++ type can guarantee that the output from the behavior_plugin is valid.

This PR approach certainly eliminates the need for Canonicalize processing, but the second advantage disappears because it returns an uncanonicalized EntityStatus type.

hakuturu583: I see no problem with Alternative Solution as a temporary alternative.
However, the values exchanged between other behavior_plugins and traffic_simulator do not behave that way, so I would like to plan to reflect my proposal in the future.
However, I think that would be a big task and future work is fine.

hakuturu583: @dmoszynski If you finished this task, please talk to me on this thread.

dmoszynski: @hakuturu583 
The alternative solution is developed and has been tested locally. If you see a need for any additional changes to this PR, please let me know :bow:, if not, please allow me to run the regress tests again :bow:

hakuturu583: OK. I start reviewing.
Please wait a bit.

simulation/behavior_tree_plugin/include/behavior_tree_plugin/action_node.hpp
```diff
@@ -84,41 +83,42 @@ class ActionNode : public BT::ActionNodeBase
...
   std::shared_ptr<hdmap_utils::HdMapUtils> hdmap_utils;
   std::shared_ptr<traffic_simulator::TrafficLightManager> traffic_light_manager;
-  std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus> entity_status;
+  std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus> canonicalized_entity_status;
   double current_time;
   double step_time;
   double default_matching_distance_for_lanelet_pose_calculation;
   std::optional<double> target_speed;
-  std::shared_ptr<traffic_simulator::CanonicalizedEntityStatus> updated_status;
+  std::shared_ptr<traffic_simulator::EntityStatus> non_canonicalized_updated_status;
```
hakuturu583, line 121: I analysed the code and this variable does not appear to be used.

simulation/traffic_simulator/src/utils/distance.cpp
```diff
@@ -45,6 +45,10 @@ auto lateralDistance(
   }
 }
 
+/*
+   See doc: 
+   https://github.com/tier4/scenario_simulator_v2/blob/729e4e6372cdba60e377ae097d032905b80763a9/docs/developer_guide/lane_pose_calculation/GetLongitudinalDistance.md
+*/
```
hakuturu583, line 51: Please use doxygen format.

```suggestion
/// @sa https://github.com/tier4/scenario_simulator_v2/blob/729e4e6372cdba60e377ae097d032905b80763a9/docs/developer_guide/lane_pose_calculation/GetLongitudinalDistance.md
```

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp
```diff
@@ -194,7 +190,11 @@ class EntityBase
 
   /*   */ void setOtherStatus(const std::unordered_map<std::string, CanonicalizedEntityStatus> &);
 
-  virtual auto setStatus(const CanonicalizedEntityStatus &) -> void;
+  virtual auto setStatus(const EntityStatus & status, const lanelet::Ids & lanelet_ids) -> void;
+
+  virtual auto setStatus(const EntityStatus & status) -> void;
+
+  /*   */ auto setCanonicalizedStatus(const CanonicalizedEntityStatus &) -> void;
```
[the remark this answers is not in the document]
dmoszynski: @hakuturu583 
Please have a look at the `setStatus` methods that take `EntityStatus` as object are the main reason why this PR was created. `EntityStatus` was **accepted before** by API/EntityManager, and it is not simple to only allow `CanonicalizedEntityStatus` to be accepted - **canonicalization has to happen somewhere**, while **`EntityStatus` is commonly used**...

#### Master (left unchanged)
https://github.com/tier4/scenario_simulator_v2/blob/3ed5e178adf203516bedfc1980792712eff06bfe/simulation/traffic_simulator/include/traffic_simulator/api/api.hpp#L241

For this reason, as part of this PR, 
I have developed an approach in which canonicalization **takes place in one place - in `EntityBase`**. 
Removing the `setStatus` that accepts `EntityStatus` will largely ruin this PR.

And its removal will cause that canonicalization will have to take place somewhere else, either in the `EntityManager` or in the `API`. Canonicalzation has to take place every time the position changes (e.g. msg::Pose), and it is more convenient (and safer) for it to take place in the way designed in this PR.

The primary place where canonicalization must take place is in the processing of messages from `SimpleSensorSimulator`, [here](https://github.com/tier4/scenario_simulator_v2/blob/3ed5e178adf203516bedfc1980792712eff06bfe/simulation/traffic_simulator/src/api/api.cpp#L326).
Thanks to this PR architecture, canonicalization doesn't take place here (it needs objects from `EntityBase` to do it, so it's better to do it in `EntityBase` itself).

#### This PR
https://github.com/tier4/scenario_simulator_v2/blob/3ed5e178adf203516bedfc1980792712eff06bfe/simulation/traffic_simulator/src/api/api.cpp#L326

#### Master
https://github.com/tier4/scenario_simulator_v2/blob/714dd8eda6fdd4cc190d5989610222a32be77e17/simulation/traffic_simulator/src/api/api.cpp#L133-L137

As I wrote in the description, `setCanonicalizedStatus()` which accepts `CanonicalizedEntityStatus` as an argument is only for cases where `CanonicalizedLaneletPose` is passed from `openscenario_interpreter`. Removing the canonicalization of `LaneletPose`->`CanonicalizedLaneletPose` in `OpenScenarioInterpreter` is too much to be part of this PR, but I believe it is worthwhile that this process does not take place on the scenario interpreter side. 

In summary, this PR ensures that the passed `EntityStatus` to `EntityBase::setStatus()` is always correctly canonicalized to `CanonicalizedEntityStatus`.

#### This PR

https://github.com/tier4/scenario_simulator_v2/blob/f9dbfec9468f95ae51a2216f6f873539ff210e2f/simulation/traffic_simulator/src/entity/entity_base.cpp#L532-L546

https://github.com/tier4/scenario_simulator_v2/blob/f9dbfec9468f95ae51a2216f6f873539ff210e2f/simulation/traffic_simulator/src/data_type/entity_status.cpp#L56-L91

simulation/behavior_tree_plugin/include/behavior_tree_plugin/action_node.hpp
dmoszynski: @hakuturu583 
You're right, since the [alternative solution](https://github.com/tier4/scenario_simulator_v2/pull/1320#discussion_r1695775781) was introduced it is not used, it's a souvenir - I'm already removing it. :bow:

simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp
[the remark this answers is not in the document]
hakuturu583: Thank you.
I understand that it's necessary to narrow down the areas to be canonicalized.

hakuturu583 (commented):
LGTM, please proceed to regression test.

hakuturu583:
@dmoszynski This branch conflicts to master branch.

dmoszynski:
@hakuturu583 
Conflicts have been resolved, CI is green. :bow:

yamacir-kit (approved):

hakuturu583 (approved):
LGTM

PR #1324, opened 2024-07-22
Fix/fix eigen variable definition

hakuturu583:
# Description
Fixes: https://github.com/tier4/scenario_simulator_v2/issues/1302, https://github.com/tier4/scenario_simulator_v2/issues/1312

## Abstract
Fix the definition of eigen variable for calculation of ego vehicle pose.

## Background

In https://github.com/tier4/scenario_simulator_v2/issues/1302, it is reported that kinematic state is not updated.

I have discovered a bug where pull requests from forks cannot be merged due to the Release Action malfunctioning.
Since #1313 is very important and needs to be incorporated urgently, @hakuturu583 will create a similar pull request and carry out verification and merging.

This Pull request is not @hakuturu583 's work, it is @MertClk 's work.
Thank you very much for contribution.

## References

**Reference to fix:**
- https://eigen.tuxfamily.org/dox/TopicPitfalls.html

This behavior is presumably due to the fact that the Eigen calculation is delayed by the Expression Template, resulting in undefined values. However, a tier4 internal verification showed that this part should not have been a problem even if auto had been used.
One possibility is that the behavior may be CPU architecture dependent.
The cause is unknown, but since there is no particular disadvantage, this PR is merged.

https://github.com/tier4/sim_evaluation_tools/issues/342
https://github.com/tier4/scenario_simulator_v2/pull/1313


hakuturu583:
Fixing release action is working on #1248

yamacir-kit (approved):

Issue #1342, opened 2024-08-19
Update Autoware User Interface to the latest UI


armaganarsln:
Autoware UI is updated to a more modern look and for sometime the only problematic UI we have is Scenario Simulator UI which still uses the old UI as shown below:

I am expecting Scenario Simulator UI to be updated to the Autoware UI as soon as possible.

If it's possible for @KhalilSelyan to update please let him know.

Autoware UI - Scenario Simulator UI

![image](https://github.com/user-attachments/assets/73416cb7-ac3a-4217-968d-88dee290d637)
![image](https://github.com/user-attachments/assets/b474add6-0e59-4a82-a23b-384a6ab84bc1)

armaganarsln:
@KhalilSelyan we have discussed with Mitsudome san and he recommended you to create a PR at scenario simulator and update.  If you have any issues please consult with Berkay.

hakuturu583:
> https://autowarefoundation.github.io/autoware-documentation/main/tutorials/scenario-simulation/planning-simulation/scenario-test-simulation/
> 
> 🔽
> 
> https://github.com/tier4/scenario_simulator_v2/blob/1d219827a28b82a5d0359c0918862e6b0a7c4d2a/test_runner/scenario_test_runner/launch/scenario_test_runner.launch.py#L235-L236
> 
> 🔽
> 
> https://github.com/tier4/scenario_simulator_v2/blob/1d219827a28b82a5d0359c0918862e6b0a7c4d2a/simulation/traffic_simulator/config/scenario_simulator_v2.rviz
> 
> This .rviz file needs to be updated.

Yes. it needs to be updated.

HansRobo:
We updated rviz config in https://github.com/tier4/scenario_simulator_v2/pull/1501 based on latest autoware.rviz in `autoware_launch` repository.

PR #1349, opened 2024-08-21
style: update autoware user interface to the latest ui

KhalilSelyan:
# Description
Updated the .rviz config file to include the correct layout 

### Related issue
- #1342

### DIFF
--- a/simulation/traffic_simulator/config/scenario_simulator_v2.rviz
+++ b/simulation/traffic_simulator/config/scenario_simulator_v2.rviz
@@ -4,12 +4,9 @@ Panels:
     Name: Displays
     Property Tree Widget:
       Expanded:
-        - /Map1/Lanelet2VectorMap1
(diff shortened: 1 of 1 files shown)


brkay54:
Thank you for your work. I can say that the configuration appears to be the same as PSim's configuration.
Screenshot from 2024-08-22 14-21-18

hakuturu583:
Thank you for sending Pull request!
But...

The rviz plugin added in this PR has its source code within Autoware, and to incorporate this PR, it would be necessary to add Autoware as a build dependency.
However, scenario_simulator_v2 is intentionally structured to depend only on the released packages of Autoware in order to minimize its dependency on Autoware.

hakuturu583:
We are carrying out the release process in this Issue (https://github.com/autowarefoundation/autoware/issues/5076).
We plan to contribute there, release the rviz plugin, and then incorporate this PR.
I apologize for the time this is taking.

hakuturu583:
Thank you for sending PR.
Contents in this PR was merged in #1501

PR #1351, opened 2024-08-22
use workflow dispatch in docker build

hakuturu583:
# Description

## Abstract

Reduced the execution time of Release workflow.

## Background

Release workflow included Docker Build step, which was blocking numerous PRs for longer than necessary.

## Details

Separated Release workflow and Docker workflow.

- The Docker workflow is executed through the workflow_dispatch API from within the Release workflow only when the Release workflow runs on the master branch immediately after merge. This triggers the Docker Build and image push.
- Since workflow_dispatch can also be executed directly from the GitHub UI, it is now possible to manually trigger it in case the Docker Image is not updated due to some error. See also [GitHub official document.](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow)

## References

[The repository has been forked ](https://github.com/merge-queue-testing/scenario_simulator_v2)and the modifications have been confirmed to work.

The Docker workflow is failing, but this is because it includes a step to push the image to the GitHub Container Registry of scenario_simulator_v2, and it is deliberately failing as the correct token cannot be provided.
It has been confirmed in the current Release workflow that it works when the correct token is passed.

https://github.com/merge-queue-testing/scenario_simulator_v2/pull/1

https://github.com/merge-queue-testing/scenario_simulator_v2/actions/runs/10553938073/job/29235018312

# Destructive Changes

N/A

# Known Limitations

N/A


.github/workflows/Docker.yaml
```diff
@@ -0,0 +1,80 @@
+name: Docker
+on:
+  pull_request:
+    paths:
+      - "**"
+      - "!docs/**"
+      - "!README.md"
+      - "!CONTRIBUTING.md"
```
HansRobo, line 8:
```suggestion
      - "!CONTRIBUTING.md"
      - "!.github/**"
```
Without this line, there is no point in explicitly specifying: ".github/workflows/Docker.yaml"
See also [previous workflow](https://github.com/tier4/scenario_simulator_v2/pull/1317/files#diff-c3f589b0287668535695b316897f85ac93aaf209dd7ccabc318f9c9d6846b52dL13-L14).

.github/workflows/Docker.yaml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          tool-cache: false
+
+      - name: Install docker for ubuntu runner
+        uses: docker/setup-buildx-action@v3
+
+      - name: Install QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - uses: actions/checkout@v3
```
HansRobo, line 44: This is a good opportunity to use the latest action.
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/Docker.yaml
```diff
@@ -0,0 +1,80 @@
...
+          workdir: .
+          set: |
+            *.cache-to=type=gha,mode=max
+            *.cache-from=type=gha
+          push: false
+          targets: |
+            ${{ matrix.rosdistro }}_base_${{ matrix.arch }}
+
+      - name: Build and push (${{ matrix.arch }})
+        if: github.event_name == 'workflow_dispatch'
```
HansRobo, line 68: Is this for debug?

.github/workflows/Release.yaml
```diff
@@ -142,50 +142,14 @@ jobs:
...
-          set: |
-            *.cache-to=type=gha,mode=max
-            *.cache-from=type=gha
-            *.tags=ghcr.io/tier4/scenario_simulator_v2:humble-${{ needs.release.outputs.new_version }}
-          push: ${{ github.event.pull_request.merged == true }}
-          targets: |
-            ${{ matrix.rosdistro }}_base_${{ matrix.arch }}
+      - name: Kick docker build action
+        if: github.event.pull_request.merged == true
+        run: |
```
HansRobo, line 148: Does this step really work?
What this step is calling is a trigger called repository_dispatch, and it looks like there is no trigger in Docker.yaml.

And, are there any restrictions that I have to call it via curl?
It seems simpler to call Docker.yaml with `uses` like below.

```yaml
      - name: Kick docker build action
        if: github.event.pull_request.merged == true
        uses: ./.github/workflows/Docker.yaml
```

.github/workflows/Release.yaml
```diff
@@ -142,50 +142,14 @@ jobs:
...
+          curl -L \
+            -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: token ${{ secrets.BLOOM_GITHUB_TOKEN }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/merge-queue-testing/scenario_simulator_v2/actions/workflows/Docker.yaml/dispatches \
+            -d '{"ref":"master","inputs":{"version":"3.5.2"}}'
```
HansRobo, line 155: Is this for debug?
Please make it a variable, not a magic number, so it works in production.

.github/workflows/Release.yaml
HansRobo: I understand that this step does not triggers `repository_dispatch` but `workflow_dispatch`.

.github/workflows/Docker.yaml
HansRobo: Sorry, this is my misunderstanding.
https://github.com/tier4/scenario_simulator_v2/pull/1351#discussion_r1728663124

.github/workflows/Release.yaml
hakuturu583: Sorry, it was my mistake.
I fix.

hakuturu583: This line was fixed.

HansRobo: > **Please use production repository**  and use a variable, not a magic number, so it works in production.

There does not seem to be fixed about the bold part of the comment.

HansRobo (approved):

PR #1303, opened 2024-06-26
Customize API::getGoalPoses for goal modifications.

TakanoTaiga:
# Description

## Abstract

This pull request customizes the API::getGoalPoses function for goal modifications in Autoware.

## Background

The goal position in Autoware is automatically modified by a feature called "goal modification." This PR addresses the need to retrieve the actual goal positions considering these modifications.

## Details

The `API::getGoalPoses` function has been customized to check the goal field in `/planning/mission_planning/route` and retrieve the goal positions. When `API::getGoalPoses` is called for an Ego entity, this new logic is applied to extract the correct goal positions. Currently, `VehicleEntity::getGoalPoses` is being called, which only computes the route to the goal without considering the goal modifications.

## References

N/A

# Destructive Changes

N/A

# Known Limitations

N/A

PR #1385, opened 2024-09-17
Customize API::getGoalPoses for goal modifications.

TakanoTaiga:
# Description

## Abstract

⚠️This PR has been resubmitted due to a conflict with #1303. There are some changes included.
This pull request customizes the API::getGoalPoses function for goal modifications in Autoware.

## Background

The goal position in Autoware is automatically modified by a feature called "goal modification." This PR addresses the need to retrieve the actual goal positions considering these modifications.

## Details

The `API::getGoalPoses` function has been customized to check the goal field in `/planning/mission_planning/route` and retrieve the goal positions. When `API::getGoalPoses` is called for an Ego entity, this new logic is applied to extract the correct goal positions. Currently, `VehicleEntity::getGoalPoses` is being called, which only computes the route to the goal without considering the goal modifications.

## References

https://github.com/tier4/sim_evaluation_tools/issues/351

# Destructive Changes

N/A

# Known Limitations

N/A


pyproject.toml
```diff
@@ -25,5 +25,5 @@ catkin-pkg = "^1.0.0"
 
 
 [build-system]
-requires = ["poetry-core"]
+requires = ["poetry-core", "setuptools"]
```
hakuturu583, line 28: Why do you update this line?

hakuturu583 (changes requested):

pyproject.toml
TakanoTaiga: I checked and found that the code was unnecessary, so I deleted it.

simulation/traffic_simulator/src/entity/ego_entity.cpp
```diff
@@ -100,6 +100,21 @@ auto EgoEntity::getBehaviorParameter() const -> traffic_simulator_msgs::msg::Beh
...
+  std::vector<CanonicalizedLaneletPose> lanelet_pose;
+
+  if (const auto universe =
+        dynamic_cast<concealer::FieldOperatorApplicationFor<concealer::AutowareUniverse> *>(
+          field_operator_application.get());
+      universe) {
+    lanelet_pose.push_back(traffic_simulator::pose::toCanonicalizedLaneletPose(
+                             universe->getAutowareMissionRoute().goal_pose, false, hdmap_utils_ptr_)
+                             .value());
+  }
```
hakuturu583, line 114: Please check optional value.

hakuturu583 (changes requested):

PR #1396, opened 2024-09-26
Use fmt library for string formatting

f0reachARR:
# Description

## Abstract

In this PR, I will introduce `fmt` formatting library instead.
This will make JSON serialization faster.

## Background

Currently, we use `stringstream` to format string, which is used to be a part of Context JSON.
However, `stringstream` is known as "slower" implementation, and this is one of bottleneck in JSON serialization.

> image
> Cite from [fmt](https://github.com/fmtlib/fmt)

### DIFF
--- a/openscenario/openscenario_interpreter/package.xml
+++ b/openscenario/openscenario_interpreter/package.xml
@@ -11,6 +11,7 @@
 
   <depend>concealer</depend>
   <depend>geometry_msgs</depend>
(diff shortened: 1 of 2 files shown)


f0reachARR:
In some my experiments, stringstream is faster than fmt. I think something is strange.
I will investigate it.

HansRobo:
This is marked as merged because all commits are included in #1418

Issue #1393, opened 2024-09-25
EntityBase “requestSpeedChangeWithConstantAcceleration” and “requestSpeedChange” sometimes will not throw an error


gmajrobotec:
**Description**

Location: enetity_base.cpp: 305, 394, 457

Function `isTargetSpeedReached` throws an error from speed_change.cpp:47 if it is provided with an invalid entity name in the `speed_change::RelativeTargetSpeed` argument. 
There are several places in `requestSpeedChange` functions where `isTargetSpeedReached` is invoked if-and-only-if `continuous == false` (early return). This means that an invalid `speed_change::RelativeTargetSpeed` argument will sometimes result in an error immediately, but sometimes will not.
Note that the error will be thrown eventually anyway (might remove this, as it is not technically an issue).

**Solution**
This could be changed by inverting the order of statements in the `if (!continuous && isTargetSpeedReached(target_speed))` statement. However as noted above, this is only a suggestion and can be removed.

hakuturu583:
Invalid operation should not be permitted, so if the `speed_change::RelativeTargetSpeed` called with invalid name, it should throw exception.

PR #1404, opened 2024-10-01
RJD-1336/fix_request_speed_change_throws

robomic:
## Abstract

Please refer to [this issue](https://github.com/tier4/scenario_simulator_v2/issues/1393).

## Details

1. Inversion of order in if statements

`  if (!continuous && isTargetSpeedReached(target_speed)) `  --->
 ` if (isTargetSpeedReached(target_speed) && !continuous)`
this forces calculation of relative_target_speed, which throws immediately, if invalid.
2. Numerous `const` qualifiers added.
3. A single `cpp_mock_scenario` has been modified in order to update `other_status` variable.
## References
[issue](https://github.com/tier4/scenario_simulator_v2/issues/1393)
[jira ticket](https://tier4.atlassian.net/browse/RJD-1336)


robomic:
[Regression results](https://tier4.atlassian.net/browse/RJD-1336?focusedCommentId=190062): OK

mock/cpp_mock_scenarios/src/speed_planning/request_speed_change_relative.cpp
```diff
@@ -67,6 +67,7 @@ class RequestSpeedChangeRelativeScenario : public cpp_mock_scenarios::CppScenari
         34741, 10.0, 0.0, api_.getHdmapUtils()),
       getVehicleParameters());
     api_.setLinearVelocity("front", 3);
+    api_.updateFrame();
```
hakuturu583, line 70: This function is intended to be called on the base type and is not expected to be called within scenarios. Access rights modifications will be implemented in a separate PR, so no changes are needed in this PR.

simulation/traffic_simulator/src/entity/entity_base.cpp
```diff
@@ -329,19 +330,20 @@ void EntityBase::requestSpeedChangeWithConstantAcceleration(
...
-          if (diff > 0) {
+          if (diff > 0.1) {
             setAccelerationLimit(std::abs(acceleration));
             return false;
           }
-          if (diff < 0) {
+          if (diff < 0.1) {
             setDecelerationLimit(std::abs(acceleration));
             return false;
           }
```
hakuturu583, line 349: Use constexpr for hard coded parameter 0.1

E.g 

```
constexpr double speed_difference_threshold = 0.1;
if (std::abs(diff) <= speed_difference_threshold) {
    return true;
}
if (diff > speed_difference_threshold) {
    setAccelerationLimit(std::abs(acceleration));
    return false;
}
```

hakuturu583 (changes requested):

mock/cpp_mock_scenarios/src/speed_planning/request_speed_change_relative.cpp
robomic: This line is needed in order to pass this particular test. Without it `other_status_` variable is not updated and does not contain other entities. Empty `other_status_` results in a throw when calculating the absolute speed from `speed_change::RelativeTargetSpeed` variable.

hakuturu583: If so, you should wait for one frame to elapse before making a decision, and should not implement the decision on the spot.

hakuturu583 (changes requested):

mock/cpp_mock_scenarios/src/speed_planning/request_speed_change_relative.cpp
robomic: i have moved `requestSpeedChange` to `onUpdate`

hakuturu583 (approved):
LGTM

PR #1408, opened 2024-10-02
Feature/joblist update traveled distance

TakanoTaiga:
# Description

## Abstract
This pull request removes the `updateTraveledDistance` function from several entity classes. The function has been integrated into the job list management system for better cleaner code.

## Background
The `updateTraveledDistance` function was previously implemented in multiple entity classes. However, a more efficient and centralized method to track traveled distance through the job list system was introduced. This refactor eliminates redundant code and ensures consistency across different entity types.

## Details
The following changes were made:
- Removed the `updateTraveledDistance` function from `EntityBase`, `EgoEntity`, `PedestrianEntity`, and `VehicleEntity`.
- The traveled distance is now updated through the job list system in `EntityBase` using a new job that calculates the traveled distance based on linear velocity.
- Updated unit tests by removing tests specifically for `updateTraveledDistance` as the functionality is now handled by the job system.

## References
N/A

# Destructive Changes
N/A

# Known Limitations
N/A

### DIFF
--- a/simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp
+++ b/simulation/traffic_simulator/include/traffic_simulator/entity/entity_base.hpp
@@ -224,7 +224,7 @@ public:
 
   /*   */ auto updateStandStillDuration(const double step_time) -> double;
 
-  /*   */ auto updateTraveledDistance(const double step_time) -> double;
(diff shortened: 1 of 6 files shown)


hakuturu583 (changes requested):
Code is looks good to me.
Start regression test.

hakuturu583 (approved):

Issue #1394, opened 2024-09-25
Possible square root of a negative number in LongitudinalSpeedPlanner::planConstraintsFromJerkAndTimeConstraint


gmajrobotec:
**Description**

Location: longitudinal_speed_planning.cpp:65, 74
Possible square root of a negative number. `target_speed` value can overwhelm the equation’s value.

**Example**
In the example attached below there are two `planConstraintsFromJerkAndTimeConstraint` invocations. In the first case, there is deceleration and the statement at line 69 provided to `std::sqrt` evaluates to -7. In the second case, acceleration occurs, and the statement at line 60 evaluates to -5.

```
TEST_F(LongitudinalSpeedPlannerTest, example)
{
  geometry_msgs::msg::Twist current_twist{};
  geometry_msgs::msg::Accel current_accel{};
  current_accel.linear.x = 1.0;
  const auto constraints =
    traffic_simulator_msgs::build<traffic_simulator_msgs::msg::DynamicConstraints>()
      .max_acceleration(1.0)
      .max_acceleration_rate(1.0)
      .max_deceleration(1.0)
      .max_deceleration_rate(1.0)
      .max_speed(10.0);

  const double target_speed = 5.0;
  const double acceleration_duration = 1.0;
  const double plausible_lower_bound = 0.0;
  const double plausible_upper_bound = 1e2;

  traffic_simulator_msgs::msg::DynamicConstraints new_constraints;
  {
    current_twist.linear.x = 10.0;
    const auto problematic_value =
      (constraints.max_deceleration_rate * acceleration_duration *
       constraints.max_deceleration_rate * acceleration_duration) -
      2 * constraints.max_deceleration_rate *
        (current_twist.linear.x - current_accel.linear.x * acceleration_duration - target_speed);

    EXPECT_TRUE(problematic_value >= 0);

    new_constraints = planner.planConstraintsFromJerkAndTimeConstraint(
      target_speed, current_twist, current_accel, acceleration_duration, constraints);

    EXPECT_CONSTRAINTS_BOUNDED(new_constraints, plausible_lower_bound, plausible_upper_bound);
  }
  {
    current_twist.linear.x = 1.0;

    const auto problematic_value =
      (constraints.max_acceleration_rate * acceleration_duration *
       constraints.max_acceleration_rate * acceleration_duration) +
      2 * constraints.max_acceleration_rate *
        (current_twist.linear.x + current_accel.linear.x * acceleration_duration - target_speed);

[...]

PR #1416, opened 2024-10-09
Enhanced variable value validation.

TakanoTaiga:
# Description

## Abstract

This pull request enhances the validation of floating point values in mathematical calculations. Add a safeSqrt function to prevent incorrect square root calculations.

## Background

Fixed what was pointed out in #1394.

## Details

In floating-point arithmetic, calculating the square root of negative numbers can result in undefined behavior. The safeSqrt function ensures that only non-negative inputs are processed, returning zero when the input is negative.

## References

- #1394 

## Destructive Changes

N/A

## Known Limitations

N/A


common/math/arithmetic/include/arithmetic/floating_point/comparison.hpp
```diff
@@ -16,12 +16,20 @@
...
+#include <cmath>
 #include <limits>
 
 namespace math
 {
 namespace arithmetic
 {
+
+template <typename T>
+auto safeSqrt(T a)
```
hakuturu583, line 28: Please add doxygen comment for describe safeSqrt behavior.

hakuturu583 (changes requested):

PR #1370, opened 2024-09-10
Reducing build time of `openscenario_interpreter` by utilizing precompiled headers

shouth:
# Description

## Abstract

This pull request reduces the build time for `opendscenario_interpreter` by using precompiled headers.

## Background

In the current codebase, large header files were being parsed multiple times and cause longer build time. To address this, I used Clang's `-ftime-trace` to identify the headers that were taking the most time to parse and updated `CMakeLists.txt` to ensure they are precompiled.

## Details

In my environment, the build time of `openscenario_interpreter` was reduced from 9m 21s to 7m 59s with 12 parallel builds.

## References

N/A

# Destructive Changes

N/A

# Known Limitations

N/A

PR #1418, opened 2024-10-10
Introduce Boost.JSON as a JSON serialization library

f0reachARR:
# Description
## Abstract

Use [Boost.JSON](https://github.com/boostorg/json) instead of nlohmann-json.
This will make faster `openscenario_interpreter` by reducing the cost of `publishCurrentContext`

## Background

`publishCurrentContext` is heavy method and took ~20ms in worst case.
However, this is required to visualize simulation status in our platform.
To achieve simulation with large number of entities, this is a major bottleneck.

## Details

In this PR, I will replace JSON serialization library which consume long time in `publishCurrentContext`.
Boost.JSON is faster library to JSON serialization. Also, it is easy to use and well maintained.

# Destructive Changes

This PR will change the way to writing JSON serialization operator `<<`.
But output should be same as before.

# Known Limitations
To faster serialization, we should use move semantics and same allocator to avoid copy of value.


shouth:
Hello, I am the author of the recently merged PR #1370. With my PR, `nlohmann/json.hpp` is now precompiled. Therefore, I think it would be beneficial to replace `nlohmann/json.hpp` with `boost/json.hpp` in the `openscenario/openscenario_interpreter/CMakeLists.txt` for this PR.

f0reachARR:
@shouth Thank you for advice! I fixed it and now PCH uses Boost.JSON

HansRobo (approved):
LGTM, waiting for green CIs.

PR #1420, opened 2024-10-10
Added checks for negative discriminants

TakanoTaiga:
## Description

This pull request adds checks for negative discriminants before calculating square roots in the longitudinal speed planning logic to prevent invalid operations that could occur when the discriminant is negative.

## Abstract

This pull request introduces validation for negative discriminants in the LongitudinalSpeedPlanner to prevent invalid square root calculations during acceleration and deceleration constraint planning.

## Background

In the current implementation, the code attempts to compute the square root of a discriminant that can occasionally be negative. This can lead to exceptions and undefined behavior. Adding checks before these calculations ensures stability and prevents such issues.

## Details

The changes involve checking if the discriminant is negative before performing a square root operation in both the acceleration and deceleration branches. If the discriminant is negative, a semantic error is thrown to handle the situation appropriately. This fix addresses the problem where the target speed may be too high or too low relative to the acceleration or deceleration rate, leading to negative values in the discriminant.

## References

N/A

## Destructive Changes

N/A

## Known Limitations

N/A

### DIFF
--- a/simulation/traffic_simulator/src/behavior/longitudinal_speed_planning.cpp
+++ b/simulation/traffic_simulator/src/behavior/longitudinal_speed_planning.cpp
@@ -56,23 +56,34 @@ auto LongitudinalSpeedPlanner::planConstraintsFromJerkAndTimeConstraint(
       (acceleration_duration * acceleration_duration);
   } else {
     if (isAccelerating(target_speed, current_twist)) {
+      double discriminant =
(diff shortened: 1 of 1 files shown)


hakuturu583 (changes requested):
LGTM
waiting for regression test.

hakuturu583 (approved):

PR #1428, opened 2024-10-23
Refactor curvature sampling loop to avoid using float counter

TakanoTaiga:
## Description

Refactor curvature sampling loop to avoid using float counter

## Abstract

Refactor of the curvature sampling loop in hermite_curve.cpp to avoid using a float counter, improving precision and readability.

## Background

This refactor addresses the potential issues with floating-point precision in the curvature sampling loop.

## Details

The original loop used a floating-point counter (for (double s = 0; s <= 1; s = s + 0.1)). This has been refactored to an integer-based loop with a fixed sample count of 10, which eliminates the need for floating-point arithmetic in the loop control.

## References

- https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=BUG&id=tier4_scenario_simulator_v2&open=AZINWTIYjvEq9OQMnMBA

## Destructive Changes

N/A

## Known Limitations

N/A

### DIFF
--- a/common/math/geometry/src/spline/hermite_curve.cpp
+++ b/common/math/geometry/src/spline/hermite_curve.cpp
@@ -316,7 +316,11 @@ std::pair<double, double> HermiteCurve::get2DMinMaxCurvatureValue() const
   /**
    * @brief 0.1 is a sampling resolution of the curvature
    */
-  for (double s = 0; s <= 1; s = s + 0.1) {
+  const int sample_count = 10;
+  const double resolution = 1.0 / sample_count;
+
+  for (int i = 0; i <= sample_count; ++i) {
+    double s = i * resolution;
     double curvature = get2DCurvature(s);
     curvatures.push_back(curvature);
   }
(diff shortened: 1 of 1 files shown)


common/math/geometry/src/spline/hermite_curve.cpp
```diff
@@ -316,7 +316,11 @@ std::pair<double, double> HermiteCurve::get2DMinMaxCurvatureValue() const
   /**
    * @brief 0.1 is a sampling resolution of the curvature
    */
-  for (double s = 0; s <= 1; s = s + 0.1) {
+  const int sample_count = 10;
+  const double resolution = 1.0 / sample_count;
```
hakuturu583, line 320: Use constexpr for constants.
Also, use the @note Doxygen command to record why the constant has the value it does.

hakuturu583 (changes requested):

hakuturu583 (approved):

PR #1437, opened 2024-11-06
extend for default lanelet matching distance for NPCs

HansRobo:
# Description

## Abstract

This pull-request extends the default matching distance offset for lanelet matching for ego and vehicle entities, from 1.0m to 1.5m. 

## Background

In some scenarios, ego or vehicle entities during lane change lost matching to lanelet.
This is not a matter at traffic_simulator, but for BehaviorTree actions especially `FollowFrontEntityAction`.

Normally, if you cut in front of a vehicle entity, it will slow down due to the action of `FollowFrontEntityAction`.
But it does not slow down when crossing entity losts a lanelet-matching because `FollowFrontEntityAction` don't recognize lanelet-matching-less entities as front entities.

> [!IMPORTANT]
> This lost of lanelet matching mainly comes from a lack of default matching distance.
So, I will extend it in this pull-request.

### Before

The red stopline by `FollowFrontEntityAction` is lost for a moment when ego is cutting in.

[shoulder_matching_1m.webm](https://github.com/user-attachments/assets/2b4a8cac-4b20-41e8-ba6e-1cb1a0ad8a3d)

### After

The red stopline by `FollowFrontEntityAction` is **NOT** lost for a moment when ego is cutting in.

[shoulder_matching_2m.webm](https://github.com/user-attachments/assets/5333828a-b538-4611-a458-e301d9d6f85f)

## References

[Regression Test: OK](https://github.com/tier4/sim_evaluation_tools/issues/377#issuecomment-2461555011)

# Destructive Changes

This pull-request could mean lane matching will now happen in places where it wasn't possible before, which is a good thing.

However, increasing this matching distance caused more failures in internal scenarios, especially since it allowed distance measurements in places where distance measurements in the lane coordinate system were NaN. This is why the increment is only 0.5m.

# Known Limitations


simulation/simple_sensor_simulator/src/vehicle_simulation/ego_entity_simulation.cpp
```diff
@@ -472,7 +472,7 @@ auto EgoEntitySimulation::setStatus(const traffic_simulator_msgs::msg::EntitySta
                                    vehicle_parameters.axles.front_axle.track_width,
                                    vehicle_parameters.axles.rear_axle.track_width) *
                                    0.5 +
-                                 1.0;
+                                 1.5;
```
hakuturu583, line 475: Please note in your comments that this value is a hard-coded parameter and the rationale for the value.

simulation/traffic_simulator/src/entity/vehicle_entity.cpp
```diff
@@ -71,11 +71,13 @@ auto VehicleEntity::getDefaultDynamicConstraints() const
...
 auto VehicleEntity::getDefaultMatchingDistanceForLaneletPoseCalculation() const -> double
 {
+  /// @note The lanelet matching algorithm should be equivalent to the one used in
+  /// EgoEntitySimulation::setStatus
   return std::max(
            vehicle_parameters.axles.front_axle.track_width,
            vehicle_parameters.axles.rear_axle.track_width) *
            0.5 +
-         1.0;
+         1.5;
```
hakuturu583, line 80: Please note in your comments that this value is a hard-coded parameter and the rationale for the value.

hakuturu583 (changes requested):

hakuturu583 (approved):

PR #1335, opened 2024-07-31
feat(traffic_controller, api): add rviz marker for TrafficSink

dmoszynski:
# Description

This PR adds Rviz visualizations for `TrafficSink` modules from `TrafficController`.

Due to the character of the changes, regression tests probably do not need to be run.

## Abstract

In order to debug such situations as in task [RJD-1283](https://tier4.atlassian.net/browse/RJD-1283), `TrafficSink` visualizations have been added to easily verify that some `TrafficSink` is not created in an unexpected position.

## Background

In Task [RJD-1283](https://tier4.atlassian.net/browse/RJD-1283), there was an issue with the unexpected despawn of an `Entity` (in this case` Ego`). 
This was caused by creating a `TrafficSink` in the center of the route that the `Entity` was following. 
The problem arose from bugs in the lanelet - lack of required joins between them. 

## Details
- The published marker is created only once, then published during each update
- Each `TrafficSink` is visualized by a marker with the shape of a cylinder, whose radius corresponds to the `TrafficSink` radius.
- On each cylinder there is a text representing the lanelet on the basis of which it was created.
-  Note: Each `TrafficSinks` is created using lanelets which do not have “Next lanelet” assigned).

Screenshot from 2024-07-31 15-33-07

## References

Jira ticket: [RJD-1283](https://tier4.atlassian.net/browse/RJD-1283)
Regressions: [jira comment](https://tier4.atlassian.net/browse/RJD-1283?focusedCommentId=181693)

[RJD-1283]: https://tier4.atlassian.net/browse/RJD-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RJD-1283]: https://tier4.atlassian.net/browse/RJD-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RJD-1283]: https://tier4.atlassian.net/browse/RJD-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1443, opened 2024-11-08
Feature/enable specify entity type in autosink

hakuturu583:
# Description

## Abstract

Enabled selection of Entity to be Sinked in TrafficSink.

## Background

In TrafficSink, arbitrary Entity was despawned when it approached TrafficSink, so there were cases where EgoEntity was despawned and testing could not continue.

## Details

To enable this feature, set auto_sink, a member of `traffic_simulator::Configuration` structure, to true, and set `std::set<uint8_t>` to specify the target EntityType of the entity to be despawned.

e.g

```cpp
auto configuration = traffic_simulator::Configuration(map_path);
configuration.auto_sink = true;
configuration.sinkable_entity_type = {traffic_simulator::EntityType::VEHICLE}
```

If you want to check the uint_8 type number assigned to the EntityType you want to specify, please check the following URL.

https://github.com/tier4/scenario_simulator_v2/blob/master/simulation/traffic_simulator_msgs/msg/EntityType.msg

## References

N/A

# Destructive Changes

N/A

# Known Limitations

- In #1335, we found that the TrafficSink constructor was not designed well enough to be merged into the PR, but we will not fix it in this PR.
- API to install TrafficSink without auto_sink function to automatically install TrafficSink at the end of the road is not implemented.


simulation/traffic_simulator/include/traffic_simulator/api/configuration.hpp
```diff
@@ -33,6 +35,8 @@ struct Configuration
 
   bool auto_sink = true;
 
+  std::set<traffic_simulator::EntityType, EntityTypeComparator> sinkable_entity_type = {};
```
yamacir-kit, line 38: Instead of adding a less generic type like EntityTypeComparator, change it to add comparison operator overloads to EntityType.

yamacir-kit (changes requested):

PR #1462, opened 2024-11-28
Adapt optional StartTrigger in Act (OSC 1.3)

f0reachARR:
# Description

## Abstract

Adapt optional StartTrigger in Act (OSC 1.3).

## Background

From [OSC 1.3](https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_XML/v1.3.0/generated/content/Act.html), StartTrigger in Act become optional. And default behavior is `If no trigger is defined, the act starts when the Storyboard enters runningState.`

## Details

I set default value for StartTrigger.
Also I added `always_true` constant to `Trigger`. In some case, we need Trigger that should be evaluated to be true. This constant helps implement such case(e.g. #774).

# Known Limitations

`openscenario_utility` does not use OSC 1.3 schema. Without using it, I cannot use optional StartTrigger because validation by it will fail.
In my local environment, I checked this PR works by modifying `openscenario_utility`.


HansRobo (approved):
LGTM, wih no regression

PR #1464, opened 2024-11-28
RJD-1457/traffic_sink_refactor

robomic:
## Abstract
This PR contains the continuation of https://github.com/tier4/scenario_simulator_v2/pull/1443
## Details
1. There is no longer `lanelet::Id` passed into TrafficSink, but an `optional<lanelet::Id>`. It is still used solely for the purpose of visualization, where `nullopt` means that the sink was placed too far from the lane. Resigning from passing `lanelet::Id` altogether turned out to be problematic, because ends of multiple lanelets overlap and the corresponding lanelet could not be determined.
2. Additionally I would like to propose an alternative approach to the `set sinkable_entity_types` variable. 
3. I have removed the special case `if sinkable_entity_types is empty, then all entities should sink`. In my opinion it could be misleading (someone could pass empty `sinkable_entity_types` and expect that no sinking behavior will occur). Now if we want to sink all entities, we need to provide the full set `sinkable_entity_types`.
4. I have removed the `bool auto_sink` variable, which could be equivalent to `sinkable_entity_types.empty()` with this approach.
5. The default behavior is `sinkable_entity_types = {}`, that is no auto sinks are generated.
## References
[Regressions OK](https://github.com/tier4/sim_evaluation_tools/issues/416#issuecomment-2535447212)


mock/cpp_mock_scenarios/launch/mock_test.launch.py
hakuturu583, line 1: This file is written in a way that it cannot be formatted to align line and variable heads.
If you do not need it, please modify it to the original format.

simulation/traffic_simulator/src/traffic/traffic_sink.cpp
```diff
@@ -38,59 +40,65 @@ namespace traffic_simulator
...
+auto TrafficSink::execute(
+  [[maybe_unused]] const double current_time, [[maybe_unused]] const double step_time) -> void
 {
-  const auto names = get_entity_names_function();
-  for (const auto & name : names) {
-    const auto pose = get_entity_pose_function(name);
-    if (math::geometry::getDistance(position, pose) <= radius) {
-      despawn_function(name);
+  const auto entity_names = entity_manager_ptr_->getEntityNames();
+  for (const auto & entity_name : entity_names) {
```
hakuturu583, line 52:
```suggestion
  for (const auto & entity_name : entity_manager_ptr_->getEntityNames()) {
```

hakuturu583 (changes requested):

mock/cpp_mock_scenarios/launch/mock_test.launch.py
robomic: the differences were from this [PR](https://github.com/tier4/scenario_simulator_v2/pull/1443/files); my work was based on these initial changes. I have reverted almost all format changes

hakuturu583 (approved):
LGTM

hakuturu583:
Since this PR is a continuation of #1443, I am currently checking with the reviewer of the original PR, @yamacir-kit, to see if a review by him is needed.

PR #1468, opened 2024-12-03
make Entity lane-changeable everywhere

HansRobo:
# Description

## Abstract

allow lane-changes everywhere in `VEHICLE_WITH_ROAD_SHOULDER` rouitng graph

## Background

Basically, the lanelet2 lane change permission/disallowance setting is intended for Autoware and was not designed with the behavior of entities running in the traffic_simulator in mind.

Also, when an entity is commanded to change lanes in a scenario, it is considered appropriate to change lanes, even if that command ignores the Road Traffic Act and lane change settings.

## Details

- make the linestrings between `road` and `road_shoulder` into `solid`(un-lane-changeable) from `dashed`(lane-changeable)
- allow lane-changes everywhere in `VEHICLE_WITH_ROAD_SHOULDER` rouitng graph

### details of `getLeftLaneletIds` / `getRightLaneletIds` changes

In this pull-request, I changed `lanelet2` function mapping switched by `include_oppsite_direction` flag.

**previous mappping**

- `include_oppsite_direction`: true(The de facto default)
  -  `lanelet::routing::RoutingGraph::adjacentRights/adjacentLefts`
- `include_oppsite_direction`: false
  - `lanelet::routing::RoutingGraph::rights/lefts`

**incoming mappping**

- `include_oppsite_direction`: true
  -  no mapping (throw an exception)
- `include_oppsite_direction`: false(The de facto default)
  - `lanelet::routing::RoutingGraph::rights/lefts`

> [!IMPORTANT]
> `lanelet::routing::RoutingGraph::adjacentRights/adjacentLefts` do not have the ability to acquire adjacent lanes, 
> including the opposite lane.
> The only difference between `rights` and `adjacentRights` is whether the adjacent lanelet they acquire is routable or not. 
> **(Both do not have the ability to acquire the opposite lanes.)**

image

### routing changes possibility due to this change

In this pull-request allows lane-changes everywhere.
This change will result in some changes to routing, as exemplified by the tests below.
Routing will always prioritize smaller cost (in this case, shorter distance), so longitudinal distance measurements may be smaller than before.

**Before this pull-request**
The routing graph and adjacent lane search options only allow you to get this route.
image

**After this pull-request**
By expanding the range of the RoutingGraph, we can compare more candidates and select a route with a shorter distance (in this case, the route is shorter if it goes on the inside of the curve).
image

## References

[Regression Test: OK](https://github.com/tier4/sim_evaluation_tools/issues/407)

# Destructive Changes

Please see "routing changes possibility due to this change" in `Details` section.

# Known Limitations


simulation/traffic_simulator/include/traffic_simulator/hdmap_utils/hdmap_utils.hpp
```diff
@@ -204,7 +204,7 @@ class HdMapUtils
 
   auto getLeftLaneletIds(
     const lanelet::Id, const traffic_simulator::RoutingGraphType,
-    const bool include_opposite_direction = true) const -> lanelet::Ids;
+    const bool include_opposite_direction = false) const -> lanelet::Ids;
```
hakuturu583, line 207: Changing the default argument bool may cause unintended behavior changes in C++ scenarios.
It would be desirable to eliminate the default argument and make it an error if the default argument is used, so that the user can set the behavior correctly.

hakuturu583 (changes requested):

simulation/traffic_simulator/include/traffic_simulator/hdmap_utils/hdmap_utils.hpp
HansRobo: Please check the implementation changes of `getLeftLaneletIds`.
Default behavior is not changed.

HansRobo: I fixed in [`2ebc41a` (#1468)](https://github.com/tier4/scenario_simulator_v2/pull/1468/commits/2ebc41a90f1eeb5443410c951f8b3f1654a6d54f)

hakuturu583 (approved):

PR #1490, opened 2024-12-18
Feature/is in intersection

hakuturu583:
# Description

## Abstract

Add `auto HdMapUtils::isInIntersection(const lanelet::Id lanelet_id) const -> bool` function.

## Background

In the current master branch implementation, there was no function to determine if a lanelet is a lanelet at an intersection.

## Details

Since it was found that lanelets included in an intersection have an attribute called `turn_direction`, it was decided to make a judgment based on this attribute.

The use of `intersection_area` polygon was also suggested, but the condition is that `intersection in lanelet2 means `turn_direction`, and polygon(intersection_area) is the polygon that is responsible for setting the area.
Therefore, we decided to use the presence or absence of `turn_direction` as a condition for judgment.

### Lanelet outside intersection.

Screenshot from 2024-12-18 12-58-17

### Lanelet inside intersection.

Screenshot from 2024-12-18 12-58-29

## References

[Internal link.](https://star4.slack.com/archives/C024FRX4UNL/p1734422030937639)
[Document about lanelet2(Japanese)](https://docs.pilot.auto/reference-design/common/map-requirements/vector-map-requirements/category_intersection/#vm-03-02-%E3%83%AC%E3%83%BC%E3%83%B3%E3%81%AE-turn_direction-%E3%81%A8-virtual)
[Document about lanelet2(English)](https://github.com/autowarefoundation/autoware_common/blob/main/tmp/lanelet2_extension/docs/lanelet2_format_extension.md#turn-directions)

# Destructive Changes

N/A

# Known Limitations

N/A


simulation/traffic_simulator/src/hdmap_utils/hdmap_utils.cpp
```diff
@@ -1376,6 +1377,11 @@ auto HdMapUtils::getVectorFromPose(const geometry_msgs::msg::Pose & pose, const
   return vector;
 }
 
+auto HdMapUtils::isInIntersection(const lanelet::Id lanelet_id) const -> bool
+{
+  return lanelet_map_ptr_->laneletLayer.get(lanelet_id).hasAttribute("turn_direction");
```
dmoszynski, line 1382: In my feeling, using `“turn_direction”` to simply get information about whether a lanelet is at an intersection is most proper. 

However, I would (at least) add a comment to this method about the fact that `“turn_direction”` **must be set** in the lanelet that is at the intersection, but **can be set** on other lanelets. As can be read [here](https://github.com/autowarefoundation/autoware_common/blob/main/tmp/lanelet2_extension/docs/lanelet2_format_extension.md#turn-directions):

> Users must add "turn_direction" tags to lanelets within intersections to indicate vehicle's turning direction. You do not need this tags for lanelets that are not in intersections.

Of course, I realize that in practice it is almost never set, but it can happen that a lanelet that is not at an intersection has `turn_direction==“straight”` set - in that case this method will return true for that lanelet, which is incorrect and can be confusing. Let me know whether you think the same. :bow: 

Also, I would consider how to handle crosswalks here, since they never have a `“turn_direction”`. Maybe it would make sense to return `“true”` for each pedestrian crosswalk, since it by its nature always intersects with the road? :bow:

hakuturu583: Sorry for sending [Japanese documentation](https://docs.pilot.auto/reference-design/common/map-requirements/vector-map-requirements/category_intersection/#vm-03-02-%E3%83%AC%E3%83%BC%E3%83%B3%E3%81%AE-turn_direction-%E3%81%A8-virtual), but lanelet2 which used by Autoware in intersection always has "turn_direction" attribute.

yamacir-kit (approved):

PR #1499, opened 2024-12-27
fix(traffic_light): don't call addTrafficLight() when the traffic light id has already existed in map

satoshi-ota:
# Description

The following function in `traffic_light_base.cpp` is quite heavy because it always tries adding traffic light even when the traffic light has been registered in simulator.

```c++
auto TrafficLightsBase::getTrafficLight(const lanelet::Id traffic_light_id) -> TrafficLight &
{
  addTrafficLight(traffic_light_id);
  return traffic_lights_map_.at(traffic_light_id);
}
```

As a result, it becomes a bottle neck of simulator when I write a code to update traffic light color frequently.

Screenshot from 2024-11-13 13-29-45

## Abstract

I added the guard to prevent redundant process like this:

```patch
diff --git a/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp b/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
index 48110784a..906e80258 100644
--- a/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
+++ b/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
@@ -123,6 +123,10 @@ auto TrafficLightsBase::addTrafficLight(const lanelet::Id traffic_light_id) -> v
 
 auto TrafficLightsBase::getTrafficLight(const lanelet::Id traffic_light_id) -> TrafficLight &
 {
+  if (isTrafficLightAdded(traffic_light_id)) {
+    return traffic_lights_map_.at(traffic_light_id);
+  }
+
   addTrafficLight(traffic_light_id);
   return traffic_lights_map_.at(traffic_light_id);
 }

```

## Background

N/A

## Details

With this PR, I confirmed the process load became light.

image

## References

N/A

# Destructive Changes

N/A

# Known Limitations

N/A

### DIFF
--- a/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
+++ b/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
@@ -115,6 +115,10 @@ auto TrafficLightsBase::isTrafficLightAdded(const lanelet::Id traffic_light_id) 
 
 auto TrafficLightsBase::addTrafficLight(const lanelet::Id traffic_light_id) -> void
 {
+  if (isTrafficLightAdded(traffic_light_id)) {
+    return;
+  }
+
   // emplace will not modify the map if the key already exists
   traffic_lights_map_.emplace(
     std::piecewise_construct, std::forward_as_tuple(traffic_light_id),
--- a/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
+++ b/simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
@@ -115,10 +115,6 @@ auto TrafficLightsBase::isTrafficLightAdded(const lanelet::Id traffic_light_id) 
 
 auto TrafficLightsBase::addTrafficLight(const lanelet::Id traffic_light_id) -> void
 {
-  if (isTrafficLightAdded(traffic_light_id)) {
-    return;
-  }
-
   // emplace will not modify the map if the key already exists
   traffic_lights_map_.emplace(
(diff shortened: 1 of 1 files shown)


simulation/traffic_simulator/src/traffic_lights/traffic_lights_base.cpp
```diff
@@ -123,6 +123,10 @@ auto TrafficLightsBase::addTrafficLight(const lanelet::Id traffic_light_id) -> v
 
 auto TrafficLightsBase::getTrafficLight(const lanelet::Id traffic_light_id) -> TrafficLight &
 {
+  if (isTrafficLightAdded(traffic_light_id)) {
+    return traffic_lights_map_.at(traffic_light_id);
+  }
+
   addTrafficLight(traffic_light_id);
   return traffic_lights_map_.at(traffic_light_id);
```
HansRobo, line 131: Thank you for contributing!
This may be better.
```suggestion
  if (not isTrafficLightAdded(traffic_light_id)) {
    addTrafficLight(traffic_light_id);
  }
  return traffic_lights_map_.at(traffic_light_id);
```

satoshi-ota: Make sense. I fixed it in https://github.com/tier4/scenario_simulator_v2/pull/1499/commits/1e5d510785f6117235bd9231c4b3ba2fc62f1328.

HansRobo (approved):

PR #1503, opened 2025-01-08
Feature/publisher with customizable randomizer

yamacir-kit:
# Description

## Abstract

This pull request adds the ability to randomize the value of any field in messages published by the `concealer` publisher.

Additionally, a launch argument `parameter_file_path` has been added to `scenario_test_runner` that accepts the path to the ROS 2 parameter file to specify the parameters to be given to the randomizer.

## Background

Previously, scenario_simulator_v2 could randomize some of the perception-related messages it published to Autoware (which is the responsibility of `simple_sensor_simulator`), but not the localization-related ones (which is the responsibility of `concealer`).

Recently, the Autoware development team requested the ability to randomize odometry out of localization, so we have implemented this and added a comprehensive parameter file to specify the parameters to be given to such a functionality.

## Details

 In this pull request, the randomizer is only enabled for those related to odometry.

The Localization topic randomizer can be configured from the ROS 2 parameter file given to the launch argument `parameter_file_path` of scenario_test_runner. The default value of `parameter_file_path` is the path to [the sample parameter file](test_runner/scenario_test_runner/config/parameters.yaml).

The parameter file was designed with the intention that in the future it will comprehensively handle the parameters given to the `simple_sensor_simulator` and the `concealer` handling publishers and subscribers. This includes parameters related to perception, which are currently specified by Properties in the OpenSCENARIO XML scenario file. Therefore, the parameter file includes information such as `architecture_type` values, message type names, topic type names, and other information that may be needed in the future (some of which is not used in this pull request).

## References

None.

# Destructive Changes

None. 

# Known Limitations

None.


xtk8532704 (commented):
Thank you!!

HansRobo (approved):

PR #1513, opened 2025-01-23
chore: change junit inline namespace to normal namespace

HansRobo:
# Description

## Abstract

Change the `junit` inline namespace to a normal namespace to avoid `common::Error` being ambiguous.

## Background

There are two `Error` classes in `common` namespace.
- `common::junit::Error`
- `common::scenario_simulator_exception::Error`

But, the both of second level namespace, `junit` and `scenario_simulator_exception` are inline namespace.
Therefore, `common::Error` is a valid expression for both of `Error` classes, and in an environment where both are valid, it will cause the following compilation error:

```
 error: reference to ‘Error’ is ambiguous
  |     throw common::Error(
```

## Details

Mostly, `common::Error` means `common::scenario_simulator_exception::Error`.
Therefore, I modified `common::Error` to mean only `common::scenario_simulator_exception::Error`.

## References

[RegressionTest: WIP]()

# Destructive Changes

If external code that uses `common::Error` for `common::junit::Error` is present, it will occur compilation error.
Note: This is a possibility, and as far as I know, such a code has never actually existed.

# Known Limitations

None

### DIFF
--- a/common/simple_junit/include/simple_junit/junit5.hpp
+++ b/common/simple_junit/include/simple_junit/junit5.hpp
@@ -28,7 +28,7 @@ using string = std::string;
 
 namespace common
 {
-inline namespace junit
+namespace junit
 {
 struct Pass
 {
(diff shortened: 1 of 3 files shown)


HansRobo:
Since I have found that this name collision was intentional by the author ( @yamacir-kit ) , I am closing this PR.
The author's current position is that the full name of the `Error` class should basically be used, and that it is only possible to omit it when the intermediate namespace can be omitted.
Therefore, in an environment with mixed definitions, where an error occurs stating that `common::Error` is ambiguous, the abbreviated name `common::Error` is likely to confuse not only compilers but also humans, so it is better to use the full name and make it clear which `Error` class it is.

PR #1334, opened 2024-07-31
RJD-1057 (4/5): Remove non-API member functions: EntityManager’s member functions forwarded to EntityBase (2/2)

TauTheLepton:
# Description

> [!IMPORTANT]
> ### Right before merging this PR, please change the target branch of #1337 to `master`.
> Otherwise that PR (which is next in line to be merged) will be closed automatically.

## Abstract

This pull request mainly consists of refactoring and introduces changes in the `API` and `EntityManager`.
This pull request was divided into some parts (#1552, #1543, #1540, #1539) to make it easier to review.

## Background

This pull request is an extension of the #1473.
Due to the fact that these pull requests combined would have been too large for a review, they have been split into two.

As a result of the previous PR, changes have been made to the `API` and `EntityManager` classes.
These changes created a great opportunity to declutter / tidy up the code of these classes.
This pull request consists of the code cleanup and refactoring. **Many of the changes introduced are just changes to the order of functions** inside the implementation file to keep them grouped by functionality and consistent with the header file.

> [!TIP]
> When reviewing the changes, it may be easier to see the actual modifications by opening both versions of each file side by side rather than looking at the diff generated by GitHub.
> This is due to the fact that GitHub shows lots of changes where mostly copying and pasting the code has been done (for reordering purposes).

### Cpp mock scenarios
The change mentioned above has been applied in the whole codebase, **which includes most of the cpp mock scenarios**.

Cpp mock scenarios have been all changed like the following:

Before:
https://github.com/tier4/scenario_simulator_v2/blob/3592d9075178b5543ac4ff3803ccadad4e1ec1d3/mock/cpp_mock_scenarios/src/follow_front_entity/accelerate_and_follow.cpp#L65-L70

After:
https://github.com/tier4/scenario_simulator_v2/blob/b3f72359256e54bf607b925dfb35cbe494b55092/mock/cpp_mock_scenarios/src/follow_front_entity/accelerate_and_follow.cpp#L65-L69

## Details

Below are all changes made in this PR

### [API](https://github.com/tier4/scenario_simulator_v2/pull/1334/files#diff-857e52e388cf27e5a57270af148fee7c7fb69fcbf9ae712e08e44183f2a90b2e)

- ~~Renamed member `configuration` to `configuration_`~~
  - Merged by #1540
- ~~Most of the constructor logic has been moved to `API::init()` member function~~
  - Merged by #1540
- Added member function `setSiulationStepTime`
[...]

PR #1543, opened 2025-03-06
RJD-1057: Unify spawn API

f0reachARR:
# Description

## Abstract

This is a partial pull request of #1334.
Unify multiple `spawn`  API to one `spawn` API

## Background

There are three `spawn` API for each entity types.
This PR will unify them into one `spawn` API.
Also, `spawn` will return Entity instead of boolean of spawn result.

## References
#1334

# Known Limitations

Returned Entity is not used now. I will introduce it by another PR.


hakuturu583 (approved):
LGTM

PR #1555, opened 2025-04-01
Remove Autoware trajectory subscription

TauTheLepton:
# Description

## Abstract

This PR removes the subscription to a topic of type `tier4_planning_msgs::msg::Trajectory`.

## Background

This subscription was present in the code since the beginning of scenario simulator. However, it is not a part of Autoware API, so ideally it should not be used.

## Details

The subscription was only used to draw markers in RViz using `EgoEntity::getWaypoints` overload.
Although, since the Autoware trajectory is already being drawn in another part of [RViz configuration](https://github.com/tier4/scenario_simulator_v2/blob/b19a88811840edc1f196da233d9729494300a9cf/simulation/traffic_simulator/config/scenario_simulator_v2.rviz#L1003-L1049), the additional visualization performed by scenario simulator is not necessary.

It was concluded, that the trajectory subscription can be safely removed, as well as the `FieldOperatorApplication::getWaypoints` member function.
To account for the deletion, the `EgoEntity::getWaypoints` member function was modified to return an empty waypoints array. An appropriate comment has been added.
This way, the visualization component will not draw any markers for `EgoEntity`'s trajectory.

### Comparison before and after the subscription removal
| Trajectory drawn for Ego by scenario simulator's visualization component | Trajectory drawn for Ego directly by the RViz trajectory plugin (after the subscription removal) |
| --- | --- |
| Screenshot from 2025-03-26 16-15-58 | Screenshot from 2025-03-27 10-18-18 |

## References

[INTERNAL LINK](https://tier4.atlassian.net/browse/RJD-1667)

# Destructive Changes

None

# Known Limitations

The new implementation of `EgoEntity::getWaypoints` always returns an empty array. This is fine for visualization, but if this member function was ever used for something different, it should be noted that `EgoEntity`'s overload of `EntityBase::getWaypoints` does not work the same as for other non-stationary entity types.
It is not possible to obtain the actual trajectory waypoints of Autoware.


TauTheLepton:
Regression tests [result](https://github.com/tier4/sim_evaluation_tools/issues/487#issuecomment-2769897212)

PR #1570, opened 2025-04-13
Add find_package(zmpqq_vendor) to fix zmqpp/zmqpp.hpp not found error

jerry73204:
When trying to build `simulation_interface` package, it fails on `zmqpp/zmqpp.hpp` not found error. It can be reproduced in the official Docker image.

```sh
rocker --volume $PWD:/monut -- ghcr.io/autowarefoundation/autoware:universe-devel-cuda-20250205-amd64
cd /mount
colcon build
```

It is fixed by adding `find_package(zmqpp_vendor REQUIRED)` to the `CMakeLists.txt` file.

### DIFF
--- a/simulation/simulation_interface/CMakeLists.txt
+++ b/simulation/simulation_interface/CMakeLists.txt
@@ -19,6 +19,7 @@ endif()
 find_package(ament_cmake_auto REQUIRED)
 find_package(Boost REQUIRED thread)
 find_package(rclcpp REQUIRED)
+find_package(zmqpp_vendor REQUIRED)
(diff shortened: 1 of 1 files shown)


hakuturu583:
Thank you for the pull request.
We believe the problem is probably caused by the zmqpp_vendor package.

Once the source code of the above package is imported into scenario_simulator_v2, fixed, and stable, I will return it to the [original repository](https://github.com/tier4/zmqpp_vendor/blob/main/CMakeLists. txt) once it becomes stable.

hakuturu583:
See also #1575

PR #1580, opened 2025-04-16
fix(openscenario_interpreter): Prevent data race in callback and get()

SzymonParapura:
# Description

This PR resolves a data race in MagicSubscription<T>, which occurred when one thread accessed the internal data (e.g., by casting a UserDefinedValue to double) while another thread was concurrently receiving and writing new data from a ROS 2 topic. The fix introduces synchronization using std::mutex, ensuring safe access to the shared state.

## Abstract

Fix data race in `MagicSubscription<T>::get()` and subscription callback by introducing mutex-based synchronization.

## Details

- Previously, get() returned data via static_cast<const T&>(*this) without locking, which could race with the callback.
- The callback copied data into the object via static_cast<T&>(*this) = *message, but it could still race with concurrent get().
- This PR introduces a std::mutex data_mutex and locks it in both the callback and get().

## References

Jira: [link](https://tier4.atlassian.net/browse/RJD-1679)

# Destructive Changes

N/A

# Known Limitations

N/A

### DIFF
--- a/openscenario/openscenario_interpreter/src/syntax/user_defined_value_condition.cpp
+++ b/openscenario/openscenario_interpreter/src/syntax/user_defined_value_condition.cpp
@@ -88,8 +88,7 @@ struct MagicSubscription : private T
     }
 
     subscription = node->template create_subscription<T>(
-      topic_name, rclcpp::QoS(1).best_effort(),
-      [this](const typename T::SharedPtr message) {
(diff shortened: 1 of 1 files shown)


SzymonParapura:
No regression [https://github.com/tier4/sim_evaluation_tools/issues/504](url)

HansRobo (approved):

PR #1337, opened 2024-08-01
RJD-1057 (5/5): Remove non-API member functions: Improve responsibility simulator_core<->api

TauTheLepton:
# Description

## Abstract

This pull request aims to improve the responsibility division between `traffic_simulator` API and `openscenario_interpreter` SimulatorCore.
The changes include refactoring, renaming functions and reordering them in the source files. So even though the number of changes reported by github is rather high, many of the changes consist of moving functions around with slight if none modifications.

## Background

This pull request is one of many that aim to modularize the scenario_simulator_v2.

### Canonicalization
The main changes are that canonicalization process (of lanelet pose) is a responsibility of `traffic_simulator`. This way the communication with `traffic_simulator` API is done on **NON** canonicalized data structures and traffic_simulator has to canonicalize the data.

> [!NOTE]
> A result of aforementioned changes, most of `cpp_mock_scenarios` had to be adjusted to communicate using regular lanelet pose instead of canonicalized one.
> This leads to GitHub reporting many lines changed and the PR seem to be big, but many of these changes are simply adjusting `cpp_mock_scenarios` like the following.
>
> Before
> https://github.com/tier4/scenario_simulator_v2/blob/bcc35aea31024b26d1a5deb43699f64f7f39c9ad/mock/cpp_mock_scenarios/src/collision/crashing_npc.cpp#L54-L56
> After
> https://github.com/tier4/scenario_simulator_v2/blob/f1cb1c2704ef7ba16dd553b107ef058c29899f50/mock/cpp_mock_scenarios/src/collision/crashing_npc.cpp#L54-L56

> [!IMPORTANT]
> Because of the change of canonicalization, the NativeLanePosition type in SimulatorCore has been changed
>
> Before
> https://github.com/tier4/scenario_simulator_v2/blob/bcc35aea31024b26d1a5deb43699f64f7f39c9ad/openscenario/openscenario_interpreter/include/openscenario_interpreter/simulator_core.hpp#L37-L39
> After
> https://github.com/tier4/scenario_simulator_v2/blob/f1cb1c2704ef7ba16dd553b107ef058c29899f50/openscenario/openscenario_interpreter/include/openscenario_interpreter/simulator_core.hpp#L36-L38

### SimulatorCore
This PR focuses on improving the interaction between the API and SimulatorCore.
This is why SimulatorCore needed many modifications.
Below they are broken down into the subclasses of SimulatorCore.

#### General
[...]

PR #1581, opened 2025-04-17
RJD-1057: Small refactor for simulator API

f0reachARR:
# Description
## Abstract

Small refactor for simulator API.
Includes:

- Reduce `xs...` parameter pack usage in simulator_core.hpp for readability
- Reduce template usage in coordinate conversion
- Use outputstream helper functions for some error messages

## References

This is a partial PR of #1337.

# Destructive Changes

`convert<T>` API in  simulator_core.hpp is removed


hakuturu583 (changes requested):
The code is LGTM, but I think developer of openscenario_interpreter should check this PR.
To avoid accidental merging, leave it as `request changes`.

PR #1582, opened 2025-04-18
Fix/concealer 7

yamacir-kit:
# Description

## Abstract

If any of the states `/api/localization/initialization_state`, `/api/routing/state`, and `/api/operation_mode/state` related to Autoware state is in `UNKNOWN` state, Autoware should be considered to be initializing. However, the current `concealer::LegacyAutowareState` is not like that. This pull request fixes that bug.

## Background

This bug was introduced by #1549.

## Details

None.

## References

None.

# Destructive Changes

None. 

# Known Limitations

None.


HansRobo (approved):
Approved, waiting for green CIs

PR #1609, opened 2025-05-23
Feature/delay curventure calculation

hakuturu583:
# Description

## Abstract

Introduced a device to speed up processing of scenario_simulator_v2

## Background

There were some cases where unnecessary lane matching was being performed or where necessary lane matching might not be performed, so these were corrected.

## Details

- Reserve 128 elements while initializing `entities_` unorderd_map.
  - Changed the default reserve area of std::unoderd_map to 128 because it was found that the default reserve area of std::unoderd_map is 13 and the rehash process is run in some scenarios.
- Delay curventure calculation
  - The curvature calculation is very computationally expensive, but it is executed inside the constructor, so the process was running every time.
- Interpolate waypoints before follow trajectory.
  - The waypoint of FollowTrajcetroyAction was conventionally set to RoutePlanner without route completion, so there was a possibility that an unintended route was passed to RoutePlanner.
  - Even when a waypoint section is a short distance, lane matching processing is performed more than necessary, reducing calculation efficiency.
- `enable_perf` option for cpp_mock_scenarios

## References

https://github.com/tier4/sim_evaluation_tools/issues/524#issuecomment-2914476030

# Destructive Changes

N/A

# Known Limitations

If the interval of waypoints of FollowTrajectoryAction is more than 1 m, the computational cost may conversely increase.
However, it was determined that it is unavoidable that there are cases where the calculation cost increases due to the possibility of passing waypoints that are not appropriate for RoutePlanner in the conventional implementation.


common/math/geometry/src/polygon/line_segment.cpp
takanotaiga, line 1: The size of points remains constant within this function. Therefore, it's more efficient to compute it once at the beginning rather than multiple times.


```c++
auto getLineSegments(
  const std::vector<geometry_msgs::msg::Point> & points, const bool close_start_end)
  -> std::vector<LineSegment>
{
  const auto point_size = points.size();
  if (point_size <= 1) {
    return {};
  } else {
    std::vector<LineSegment> seg;
    for (size_t i = 0; i < point_size - 1; i++) {
      if (points[i] != points[i + 1]) {
        seg.emplace_back(points[i], points[i + 1]);
      }
    }
    /// @note If true, the end point(points[point_size - 1]) and start point(points[0]) was connected.
    if (close_start_end) {
      if (points[point_size - 1] != points[0]) {
        seg.emplace_back(points[point_size - 1], points[0]);
      }
    }
    return seg;
  }
}
```

common/math/geometry/src/spline/catmull_rom_spline.cpp
```diff
@@ -122,8 +122,64 @@ auto CatmullRomSpline::getTrajectory(
...
-CatmullRomSpline::CatmullRomSpline(const std::vector<geometry_msgs::msg::Point> & control_points)
-: control_points(control_points), line_segments_(getLineSegments(control_points)), total_length_(0)
+auto CatmullRomSpline::getTrajectoryPoses(
+  const double start_s, const double end_s, const double resolution) const
+  -> std::vector<geometry_msgs::msg::Pose>
+{
+  if (start_s > end_s) {
+    std::vector<geometry_msgs::msg::Pose> ret;
+    double s = start_s;
+    while (s > end_s) {
```
takanotaiga, line 132: Is there a possibility that this while loop could become infinite?

takanotaiga (commented):
The purpose of this PR is optimization. Do you have any quantitative data showing how much the performance has improved before and after applying this PR?

common/math/geometry/include/geometry/vector3/operator.hpp
```diff
@@ -131,6 +132,16 @@ auto operator+=(T & a, const U & b) -> decltype(auto)
   a.z += b.z;
   return a;
 }
+
+template <
+  typename T, typename U,
+  std::enable_if_t<std::conjunction_v<IsLikeVector3<T>, IsLikeVector3<U>>, std::nullptr_t> =
+    nullptr>
+auto operator==(T & a, const U & b) -> bool
```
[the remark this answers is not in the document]
hakuturu583: Fixe in https://github.com/tier4/scenario_simulator_v2/pull/1609/commits/1e57a225a8bed234ddd3db886372713da5cf724d

takanotaiga (approved):

PR #1633, opened 2025-06-23
Add noise model v3 for perception topic

HansRobo:
# Description

## Abstract

In this pull-request, I introduce noise v3 configuration for `/perception/object_recognition/detection/objects` topic.
The noise v3 configuration is an extension of v2 and has exactly the same noise capabilities, but now you can have multiple v3 noise configurations, allowing you to apply different noise configurations to specific entity groups.

The video below shows each detected entity is applied a different type of noise by noise v3.

https://github.com/user-attachments/assets/10d3100e-fcc9-46a8-b8f7-aaabd658fb9a

## Background

In previous noise configurations, we set one noise configuration to all detected entities.
But users want to set different configurations for defferent entity types. 

## Details

For noise configuration details, please see [Parameter.md](https://github.com/tier4/scenario_simulator_v2/pull/1633/files#diff-173ff4978f043d20a02a1403d52e0f688aa6fa5f980cdf75a0c49685ede42e61).

### workaround to use unknown name ros2 parameters

In noise v3 configurations, config name is user-defined.
So, we cannot hard-code the full name of parameters.
Without prior declaration, we cannot even know the existence of parameters read from YAML through the rclcpp API, so not being able to know the full path in advance was a major implementation limitation.
To get around this problem, I use `automatically_declare_parameters_from_overrides` option in `rclcpp::NodeOptions`.
By using this option, it becomes possible to use parameters without prior declaration, and you can find out the user-defined namespace names just by looking at the list of parameters.

## References

[Regression Test: OK](https://github.com/tier4/sim_evaluation_tools/issues/538)

# Destructive Changes

None

# Known Limitations

None


HansRobo:
@yamacir-kit 
Finally, regression test for the last commit is green.
https://github.com/tier4/sim_evaluation_tools/issues/538#issuecomment-3023050952

PR #1634, opened 2025-06-23
Use ament_vendor for zmqpp library

f0reachARR:
# Description

## Abstract

Use ament_vendor for zmqpp library, instead of ExternalProject.

## Details

This section is same as https://github.com/tier4/zmqpp_vendor/pull/9

## References

- https://github.com/tier4/zmqpp_vendor/pull/9

# Destructive Changes

- Install directory of zmqpp is changed


HansRobo (approved):
LGTM. I also checked the build include this on Autoware Evaluator.

xmfcx:
@f0reachARR @HansRobo

> Use ament_vendor for zmqpp library, instead of `ExternalProject`.

Could you explain why you've  added `ament_cmake_vendor_package` dependency with this PR?

`ament_cmake_vendor_package` depends on the deprecated `vcstool` and Prevents us from going ahead with [📣 Switching from `vcstool` to `vcs2l` #6645](https://github.com/orgs/autowarefoundation/discussions/6645).

I'm going to create a PR to switch to `ExternalProject` and remove the `ament_cmake_vendor_package` dependency.

This way the users can keep using the modern tooling.

PR #1644, opened 2025-07-09
Initialize twist/accel values to zero in DoNothingBehavior when not following trajectory

hakuturu583:
# Description

  ## Abstract

  Fixed DoNothingBehavior to initialize twist and accel values to zero when not in FollowTrajectory state, ensuring proper StandStillDuration calculations for stopped entities.

  ## Background

  Previously, when DoNothingBehavior was not following a trajectory, the twist and accel values were not being reset to zero. This caused issues with StandStillDuration calculations, as the system would incorrectly measure standstill duration even when the entity appeared to be stopped but still had non-zero
   velocity values internally.

  ## Details

  * Modified the `update` method in DoNothingBehavior plugin to explicitly set twist and accel to zero when in "do_nothing" state
  * Used geometry_msgs::build functions for consistent initialization with existing code style

  ## References

  N/A

  # Destructive Changes

  N/A

  # Known Limitations

  N/A

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/simulation/do_nothing_plugin/src/plugin.cpp
+++ b/simulation/do_nothing_plugin/src/plugin.cpp
@@ -192,7 +192,8 @@ void DoNothingBehavior::update(double current_time, double step_time)
     }
   } else {
     behavior = "do_nothing";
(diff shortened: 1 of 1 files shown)


takanotaiga (approved):
LGTM

Issue #1646, opened 2025-07-15
Error: failed validating 'False' with XsdPatternFacets


AnzhiZhang:
**Describe the bug**

The conversion script fails to convert boolean values from yaml to xosc.

**To Reproduce**

1. Run the scenario following the [Autoware manual](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/scenario-simulation/planning-simulation/random-test-simulation/),

```bash
source install/setup.bash
ros2 launch scenario_test_runner scenario_test_runner.launch.py \
  architecture_type:=awf/universe/20250130 \
  record:=false \
  scenario:='$(find-pkg-share scenario_test_runner)/scenario/sample.yaml' \
  sensor_model:=sample_sensor_kit \
  vehicle_model:=sample_vehicle
```

2. See error below.

```bash
root@SORA-WS3:/workspace# ros2 launch scenario_test_runner scenario_test_runner.launch.py   architecture_type:=awf/universe/20250130   record:=false   scenario:='$(find-pkg-share scenario_test_runner)/scenario/sample.yaml'   sensor_model:=sample_sensor_kit   vehicle_model:=sample_vehicle
[INFO] [launch]: All log files can be found below /root/.ros/log/2025-07-14-16-55-40-267993-SORA-WS3-706811
[INFO] [launch]: Default logging verbosity is set to INFO
architecture_type                   := awf/universe/20250130
autoware_launch_file                := planning_simulator.launch.xml
autoware_launch_package             := autoware_launch
consider_acceleration_by_road_slope := False
consider_pose_by_road_slope         := True
enable_perf                         := False
global_frame_rate                   := 30.0
global_real_time_factor             := 1.0
global_timeout                      := 180
initialize_duration                 := 30
launch_autoware                     := True
launch_rviz                         := False
output_directory                    := /tmp
override_parameters                 := 
parameter_file_path                 := /workspace/install/scenario_test_runner/share/scenario_test_runner/config/parameters.yaml
port                                := 5555
publish_empty_context               := False
record                              := false
record_storage_id                   := 
rviz_config                         := /workspace/install/traffic_simulator/share/traffic_simulator/config/scenario_simulator_v2.rviz
scenario                            := $(find-pkg-share scenario_test_runner)/scenario/sample.yaml
sensor_model                        := sample_sensor_kit
sigterm_timeout                     := 8
simulate_localization               := True
speed_condition                     := legacy
use_sim_time                        := False
[...]

PR #1649, opened 2025-07-16
Feature/minimal containers

dustinblack:
# Description

Updates the Dockerfiles to streamline the container build processes and drastically reduce the size of the final image. Also addresses a couple of related and prerequisite bug fixes and missing dependencies. Additionally enables build of arm64 containers in the github workflow.

## Abstract

The current containers are up to 14GB in size due to a lack of multi-stage builds and the inclusion of all build dependencies in the final containers. This large container size limits the ability to run the simulator containerized on target Automotive platforms. These changes result in final runtime containers of under 2.5GB in size.

Along with streamlining the Dockerfiles themselves, this PR:

- Adds a .dockerignore file to improve build efficiency
- Ignores an error in the traffic simulator build due to changes in the build dependencies
- ~~Fixes a boolean conversion bug when sourcing a scenario from a YAML file (likely duplicates #1648 and addresses #1646)~~
- Enables the building of arm64 images in the github workflow

## Background

Our Red Hat Performance & Scale team is working to use Autoware and the scenario simulator as a sample realistic load driver running on AutoSD and RHIVOS on representative arm64 automotive platforms. The existing available containers are large enough to exhaust the storage available on these test systems, preventing the running of the simulations on the target environments.

## References

Development and testing of these changes was assisted by AI using Gemini Pro 2.5

# Destructive Changes

It is possible that not all runtime dependencies are included in the final container for all use cases. I don't have the ability to test this or the insight to know for sure.

### DIFF
--- a/openscenario/openscenario_utility/openscenario_utility/conversion.py
+++ b/openscenario/openscenario_utility/openscenario_utility/conversion.py
@@ -139,10 +139,7 @@ def from_yaml(keyword, node):
                 #
                 # => @tag: { ... }
                 #
-                if isinstance(value, bool):
-                    result["@" + tag] = str(value).lower()
-                else:
-                    result["@" + tag] = str(value)
+                result["@" + tag] = str(value)
             else:
                 #
                 # Tag: { ... }


openscenario/openscenario_utility/openscenario_utility/conversion.py
```diff
@@ -128,7 +128,10 @@ def from_yaml(keyword, node):
                 #
                 # => @tag: { ... }
                 #
-                result["@" + tag] = str(value)
+                if isinstance(value, bool):
+                    result["@" + tag] = str(value).lower()
+                else:
+                    result["@" + tag] = str(value)
```
AnzhiZhang, line 134: I debated both solutions here in #1646, and ultimately chose to add one extra processing and conversion. Adding a boolean if-statement here, people may forget about the processing here long into the future. Adding more comments adds additional functionality to this function, so it might be better to separate them.

BTW, this modification doesn't seem to have much to do with this PR topic, and then my contribution will be lost :(

dustinblack: I'm more than happy to rebase this on another separate solution. The reason the fix is included here is that it was a blocker for me getting working builds with the new Dockerfiles and I had not yet seen that another PR was open with a proposed fix.

AnzhiZhang: Thanks, I'sd appreciate it if you are willing to rebase the solution. For reference, here is the PR addressing the fix: #1648 (we opened them very closely :). It’s likely to be merged, so please feel free to leave this partto that PR if you want.

dustinblack:
> Hello everyone any updates on this ? @dustinblack @AnzhiZhang

The latest update from me is in the conversation above with @AnzhiZhang . I rebased this PR on master after #1648 was merged, and I removed my own fix for that boolean issue from this PR. I've been testing a lot for my use case with the minimized container build for aarch64, but I've actually struggled to get an x86_64 version of the container to build (though I don't have a good environment to build that right now).

HansRobo (changes requested):

PR #1668, opened 2025-08-20
Fix observed errors in GitHub Actions

f0reachARR:
## Abstract

This PR will fix observed errors by #1657

## Details

This will fix these errors:

- Docker action is not triggered
- Release action fail on release commit (does not affect release process)
- Release body is empty


HansRobo (approved):
LGTM, waiting for green CIs

PR #1672, opened 2025-08-22
Replace quaternion-based orientation check with yaw-difference front …

takanotaiga:
## Abstract

Replace the previous “front entity” judgment that relied on the relative quaternion between self and target with a yaw-difference–based check computed from the bearing vector to the target. This makes the front detection independent of the target’s own orientation and selects the nearest entity within ±90° of the ego’s yaw.

## Background

The former implementation determined whether another entity was “in front” by computing the relative rotation (quaternion) between the ego orientation and the target orientation, then checking if the resulting yaw was within ±π/2. 

## Details

**Front determination logic**

  * Compute ego pose once: `self_pos` and `self_yaw`.
  * For each candidate entity:

    * Compute the vector from ego to the entity: `(dx, dy)`.
    * Compute the vector’s yaw (bearing): `vec_yaw = atan2(dy, dx)`.
    * Normalize the yaw difference:
      `yaw_diff = atan2(sin(vec_yaw - self_yaw), cos(vec_yaw - self_yaw))`.
    * Consider the entity “in front” if `|yaw_diff| ≤ π/2`.

## References

N/A

# Destructive Changes

N/A 

# Known Limitations

N/A


HansRobo (commented):
@takanotaiga 
The pull-request description only states what can be seen by looking at the file diffs.
Please add additional information about the intention and background knowledge behind the implementation, such as what problem there was, why it was implemented in this way, and whether there are any problems that this implementation does not solve(please write it in "Limitation").

PR #1673, opened 2025-08-25
Fix aggressive approach: stop within min distance, track gap tolerance

takanotaiga:
## Abstract

Refactors `FollowFrontEntityAction::doAction` to prevent overly aggressive approaches to the lead entity. The vehicle now **fully stops whenever the gap ≤ minimum stopping distance**, and only **accelerates when the gap ≥ a “stable distance”** (min-stop + ego length + configured margin). The requested speed is then **clamped to `[0, target_speed]`**, and obstacle output is computed once and reused.

## Background

The previous logic adjusted speed with three separate branches: accelerate when beyond a distance threshold, decelerate by `speed_step` when within stop distance, else match the front entity’s speed. That could allow creeping into an unsafe gap because it didn’t force a full stop, and it duplicated obstacle calculation in each branch. The new structure codifies two thresholds and clamps the speed, eliminating negative or excessive values and removing duplicated output calls. 

## Details

* Introduces `min_stop` and `stable_distance` and uses them to compute a single `requested` speed via a local lambda. If `adjusted <= min_stop`, the requested speed is **0.0**; if `adjusted >= stable_distance`, it becomes `front_entity_linear_velocity + speed_step`; otherwise it matches the front entity’s speed.
* Applies `std::clamp(requested, 0.0, target_speed_.value())` before updating the canonicalized status.
* Consolidates `calculateObstacle(waypoints)` into a local `obstacle` and sets outputs once, instead of recalculating in each branch.
* Scope of change: **1 file**, **\~25 additions / \~23 deletions** in `follow_front_entity_action.cpp`.

## References
N/A

## Destructive Changes
N/A

## Known Limitations
N/A


HansRobo (approved):
LGTM, wating for final regression test and pull-request description update to follow updated diffs.

Issue #1573, opened 2025-04-14
`scenario_simulator_v2` is not compatible `xmlschema>=4.0.0`


HansRobo:
**Describe the bug**

`scenario_simulator_v2` occurs an error with `xmlschema>=4.0.0` 

**To Reproduce**

1. ensure `xmlschema` version is greater than or equal to 4.0.0
2. run any scenario
3. see error like below.

```bash
[scenario_test_runner.py-1] Error: failed validating '1' with XsdPatternFacets(['[$][{][ A-Za-z0-9_\\+\\-\\*/%$\\(\\)\\.,]*[\\}]']):
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Reason: value doesn't match any pattern of ['[$][{][ A-Za-z0-9_\\+\\-\\*/%$\\(\\)\\.,]*[\\}]']
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Schema component:
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1]   <xsd:pattern xmlns:xsd="http://www.w3.org/2001/XMLSchema" value="[$][{][ A-Za-z0-9_\+\-\*/%$\(\)\.,]*[\}]" />
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Instance type: <class 'xml.etree.ElementTree.Element'>
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Instance:
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1]   <FileHeader description="Scenario where test collision conditions by type" author="abco20" date="2024-10-17T05:57:14.376Z" revMajor="1" revMinor="0" />
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Path: /OpenSCENARIO/FileHeader
[scenario_test_runner.py-1] 
```

**Expected behavior**
No errors occur on start-up of scenario execution.すｂ 

**Desktop (please complete the following information):**
 - Ubuntu Version

```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.5 LTS
Release:        22.04
Codename:       jammy
```

 - ROS 2 version
 - DDS

```bash
$ env | grep "ROS\|RMW"
ROS_VERSION=2
ROS_PYTHON_VERSION=3
ROS_LOCALHOST_ONLY=1
ROS_DISTRO=humble
[...]

PR #1677, opened 2025-08-28
Fixed POSE_IN_LANE reference

leonidaas:
# Description

Fix variable reference in sample_awsim.yaml

## Abstract

For running OpenScenario with AWSIM 2.0 the following command is required: 
```
source install/setup.bash
ros2 launch scenario_test_runner scenario_test_runner.launch.py                        \
architecture_type:=awf/universe/20250130  record:=false                                         \
scenario:='$(find-pkg-share scenario_test_runner)/scenario/sample_awsim.yaml'          \
sensor_model:=awsim_sensor_kit  vehicle_model:=sample_vehicle                          \
launch_simple_sensor_simulator:=false autoware_launch_file:="e2e_simulator.launch.xml" \
initialize_duration:=260 port:=8080
```

However this results in the following error:
```

[scenario_test_runner.py-1] Error: failed validating 'POSE_IN_LANE' with XsdPatternFacets(['[$][{][ A-Za-z0-9_\\+\\-\\*/%$\\(\\)\\.,]*[\\}]']):
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Reason: value doesn't match any pattern of ['[$][{][ A-Za-z0-9_\\+\\-\\*/%$\\(\\)\\.,]*[\\}]']
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Schema component:
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1]   <xsd:pattern xmlns:xsd="http://www.w3.org/2001/XMLSchema" value="[$][{][ A-Za-z0-9_\+\-\*/%$\(\)\.,]*[\}]" />
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Instance type: <class 'xml.etree.ElementTree.Element'>
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Instance:
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1]   <LanePosition roadId="" laneId="225" s="POSE_IN_LANE" offset="-0.1748">
[scenario_test_runner.py-1]     <Orientation type="relative" h="0" p="-0.0" r="0" />
[scenario_test_runner.py-1]   </LanePosition>
[scenario_test_runner.py-1] 
[scenario_test_runner.py-1] Path: /OpenSCENARIO/Storyboard/Init/Actions/Private[1]/PrivateAction[1]/TeleportAction/Position/LanePosition
[scenario_test_runner.py-1] 
```

I fixed the reference and the script is working fine again.

### DIFF
--- a/test_runner/scenario_test_runner/scenario/sample_awsim.yaml
+++ b/test_runner/scenario_test_runner/scenario/sample_awsim.yaml
@@ -101,7 +101,7 @@ OpenSCENARIO:
                     LanePosition:
                       roadId: ''
                       laneId: '225'
-                      s: POSE_IN_LANE
+                      s: ${POSE_IN_LANE}
                       offset: -0.1748
                       Orientation:
                         type: relative


HansRobo:
@leonidaas 
Thank you for the PR!
But, unfortunately, the error in this scenario doesn't seem to be an issue with the scenario, but rather an issue with the library as shown in #1573 .

In fact, a same error no longer occurs after downgrading xmlschema library like this.
```bash
pip3 install xmlschema==3.4.5
```
Although this is a temporary solution, we ask that you please downgrade the xmlschema for the time being.

Issue #1573 appears to have been fixed for booleans and integers, but the same fix may need to be applied to doubles as well.
Anyway, thanks for reporting with this PR. 
I don't think I would have noticed this issue without you. I'll look into it and try to find a fix.

PR #1731, opened 2025-11-05
Refactor `ActionNode` - partial remove `HdMapUtils` dependency

TauTheLepton:
# Description

## Abstract
This PR contains refactoring of `HdMapUtils` with the goal of moving functionality from this class to free functions.

## Background
This is a part of #1479 
See that PR for more background on why these changes are made.

## Details
All changes are listed in the points below.

### 1
`ActionNode::getDistanceToTargetEntity` member function has been replaced with `distance::splineDistanceToBoundingBox` function.

The functionality is not changed, some surrounding changes had to be made in order to achieve this separation:
- Added `getAltitudeRange` member function to all spline classes.
  - It returns lowest and highest altitude if given spline
- Added `lanelet_wrapper::pose::isAltitudeWithinRange` function
  - It checks whether the given altitude is within a given range

This change necessitated changing several places where `ActionNode::getDistanceToTargetEntity` was used before.

### 2
`ActionNode::getOtherEntitiesCanonicalizedEntityStatuses` member function has been added.
It returns Canonicalized entity statuses of all other entities.

### 3
Fixed mistakes where local variables and function arguments used a `_` suffix, which is a convention for member variables.
This mistake was introduced when renaming member variables in [this commit](https://github.com/tier4/scenario_simulator_v2/commit/2262581b468a9cd5e08d11042abc7f7f2d225f60).

### 4
Added check whether entity is in lanelet in `ActionNode::getFrontEntityName`.
This condition should be checked, because searching for entity in front is done on the lanelets.

### 5
Added check whether entity is in lanenet in `ActionNode::getDistanceToConflictingEntity`.
This condition should be checked, because searching for conflicting entity is done on the lanelets.

### 6
`HdMapUtils::getAltitude` was removed, because it was not used, and the functionality is in the traffic simulator utils.

### 7
Moved implementation of conflicting IDs functions from `HdMapUtils` to `lanelet_wrapper::lanelet_map` namespace.
The `HdMapUtils` forwarding member functions will be removed in the future.
- `HdMapUtils::getConflictingCrosswalkIds` => `lanelet_wrapper::lanelet_map::conflictingCrosswalkIds`
- `HdMapUtils::getConflictingLaneIds` => `lanelet_wrapper::lanelet_map::conflictingLaneIds`

### 8
`lanelet_wrapper::pose::isAltitudeMatching` has been renamed to `lanelet_wrapper::pose::isAltitudeWithinThreshold`.
This change makes it clear that a threshold is being used and to make it in line with `lanelet_wrapper::pose::isAltitudeWithinRange` naming.

The threshold `altitude_threshold` has been moved to header as global constant `ALTITUDE_THRESHOLD` to use it in both `isAltitudeWithinThreshold` and `isAltitudeWithinRange`.

## References
Description of #1479 and Abstract of #1472 for better overview of the changes.

# Destructive Changes
Some functions have been renamed, as explained in **Details**, but the codebase has been adjusted.

# Known Limitations
None

### DIFF
--- a/common/agnocast_wrapper/CHANGELOG.rst
+++ b/common/agnocast_wrapper/CHANGELOG.rst
@@ -2,14 +2,11 @@
 Changelog for package agnocast_wrapper
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-16.4.2 (2025-05-23)
+20.0.4 (2025-12-09)
 -------------------
-
-18.3.2 (2025-11-06)
--------------------
-* Merge branch 'master' into dependabot/github_actions/actions/checkout-5
-* Merge branch 'master' into dependabot/github_actions/actions/checkout-5
-* Contributors: Kotaro Yoshimoto, Taiga
+* Merge pull request `#1764 <https://github.com/tier4/scenario_simulator_v2/issues/1764>`_ from tier4/sort-changelog
+* Sort CHANGELOG.rst
+* Contributors: Kotaro Yoshimoto
 
 20.0.3 (2025-12-03)
 -------------------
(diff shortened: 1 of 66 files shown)


TauTheLepton:
No regressions confirmed [here](https://github.com/tier4/sim_evaluation_tools/issues/593#issuecomment-3520772868).

common/math/geometry/src/spline/catmull_rom_subspline.cpp
```diff
@@ -23,6 +23,15 @@ namespace geometry
 {
 double CatmullRomSubspline::getLength() const { return end_s_ - start_s_; }
 
+std::pair<double, double> CatmullRomSubspline::getAltitudeRange() const
+{
+  if (spline_) {
+    return spline_->getAltitudeRange();
```
HansRobo: This returns the altitude range of the entire spline, not the subspline segment [ `start_s_,` `end_s_` ]. This seems inconsistent with other methods like `getLength()` and `getCollisionPointIn2D()` which correctly respect the subspline bounds.

PR #1780, opened 2025-12-25
feat: add launch_system_monitor parameter

iwatake2222:
# Description

## Abstract

This PR adds a new parameter, `launch_system_monitor`.

```
simulations:
  - name: planning_sim_v2
    type: planning_sim_v2
    simulator:
      parameters:
        launch_system_monitor: "true"
```

## Background

Depending on the computing performance of the simulation execution environment, there are cases where simulation results become inconsistent or incorrect.
Autoware includes a resource monitoring module called System Monitor. By utilizing this module, we can obtain useful information for debugging such issues.
Actual case: [TIER IV INTERNAL LINK](https://star4.slack.com/archives/CJVHMJDJ7/p1765867045170619)

## Details

- Autoware has a parameter named `launch_system_monitor` that controls whether the System Monitor is launched.
This PR enables configuring this parameter from scenario_simulator.
- In planning_simulator, this parameter is disabled (false) by default, since System Monitor is not required in most cases. With this change, the default behavior remains disabled (false), allowing users to enable it only when necessary.
- This PR is not a mandatory change. If reviewers determine that this change is unnecessary, it is acceptable to close the PR.

## References

Test Results

- [TIER IV INTERNAL LINK](https://evaluation.tier4.jp/evaluation/reports/22243ff4-b8de-5716-b577-2141e7311c4d?project_id=prd_jt)
  - Execution result with this PR applied. Confirmed that the system behaves as before with System Monitor disabled.
- [TIER IV INTERNAL LINK](https://evaluation.tier4.jp/evaluation/reports/13ceddef-d589-5bab-8217-ccc2193bf848?project_id=prd_jt)
  - Execution result with this PR applied and `launch_system_monitor`=true set. Confirmed that System Monitor is enabled and that the related nodes are launched, as verified from the logs.

# Destructive Changes

None

# Known Limitations

None

### DIFF
--- a/simulation/traffic_simulator/src/entity/ego_entity.cpp
+++ b/simulation/traffic_simulator/src/entity/ego_entity.cpp
@@ -67,6 +67,7 @@ EgoEntity::EgoEntity(
       parameters.push_back("perception/enable_traffic_light:=" + std::string(architecture_type >= "awf/universe/20230906" ? "true" : "false"));
       parameters.push_back("use_sim_time:=" + std::string(common::getParameter<bool>(node_parameters, "use_sim_time", false) ? "true" : "false"));
       parameters.push_back("localization_sim_mode:=" + std::string(common::getParameter<bool>(node_parameters, "simulate_localization") ? "api" : "pose_twist_estimator"));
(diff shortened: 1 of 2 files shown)


iwatake2222:
This change is not driven by any specific requirement and is not mandatory.
It was added simply because it may be useful in certain situations.
If reviewers consider this change unnecessary, I am happy to cancel this PR.

PR #1844, opened 2026-05-13
fix steer report

piotr-zyskowski-rai:
# Description

## Abstract

Fix `SteeringReport` to publish the actual simulated steering angle instead of the commanded target angle.

## Background

The `SteeringReport` message was populated from the lateral control command:

```cpp
message.steering_tire_angle = getCommand().lateral.steering_tire_angle;
```

This means the simulator was reporting the *desired* steering angle to Autoware rather than the *actual* one. For `DELAY_STEER_*` vehicle models, the simulated steering lags behind the command due to actuator dynamics. Feeding the command back as the report effectively hid this lag from the controller, causing it to receive false state feedback. 

## Details

A new `std::atomic<float> current_steering_tire_angle` field is added to `AutowareUniverse`, following the same pattern as `current_twist`, `current_pose`, and `current_acceleration`.

Each simulation step, `EgoEntitySimulation::setAutowareStatus()` stores the actual value from the vehicle dynamics model:

```cpp
autoware->current_steering_tire_angle.store(static_cast<float>(vehicle_model_ptr_->getSteer()));
```

The `SteeringReport` timer callback then reads this value:

```cpp
message.steering_tire_angle = current_steering_tire_angle.load();
```

This ensures Autoware receives physically accurate steering feedback, including any delay/lag modelled by the vehicle simulation.

### DIFF
--- a/external/concealer/include/concealer/autoware_universe.hpp
+++ b/external/concealer/include/concealer/autoware_universe.hpp
@@ -73,6 +73,7 @@ public:
   std::atomic<geometry_msgs::msg::Accel> current_acceleration;
   std::atomic<geometry_msgs::msg::Pose>  current_pose;
   std::atomic<geometry_msgs::msg::Twist> current_twist;
+  std::atomic<double>                    current_steering_tire_angle{0.0};
   // clang-format on
 
 private:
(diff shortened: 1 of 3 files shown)


piotr-zyskowski-rai:
The regression tests passed, and [double-check tests](https://tier4.atlassian.net/browse/RJD-1999?focusedCommentId=308387) on several scenarios confirmed that the changes work as expected.

PR #1847, opened 2026-05-21
Open simulation interface

yamacir-kit:
# Description

## Abstract

Added a dependency on `OpenSimulationInterface/osi-cpp` to package `simulation_interface`.

## Background

As preparation for replacing `simulation_interface` with the ASAM Open Simulation Interface (OSI).

## Details

CMake FetchContent clones the necessary dependency repositories during the build process.

## References

- [OpenSimulationInterface/osi-cpp](https://github.com/OpenSimulationInterface/osi-cpp)

# Destructive Changes

None.

# Known Limitations

None.


HansRobo (approved):
Note: The way nanoseconds are handled in timestamps is that both the source and destination only handle decimal places in seconds, so you can substitute them directly.

